PR #5, opened 2014-08-05
Added conditional to support WP-CLI on local clones.

camb416:
WP-CLI automatically strips the require_once out of wp-config.php, but not wp-config-local.php. Adding this conditional fixes the issue when cloning a Pantheon wordpress to local.

### DIFF
--- a/wp-config-sample.php
+++ b/wp-config-sample.php
@@ -86,5 +86,5 @@ define('WP_DEBUG', false);
 if ( !defined('ABSPATH') )
 	define('ABSPATH', dirname(__FILE__) . '/');
 
-/** Sets up WordPress vars and included files. */
-require_once(ABSPATH . 'wp-settings.php');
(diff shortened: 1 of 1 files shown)


jhabdas:
@camb416 I believe it's safe to say this isn't going to be merged anytime soon. Would you mind closing this pull and consider opening an issue for discussion instead?

nullvariable:
This issue can be closed as the `wp-config-local` file is pulled in as a part of our wp-config defaults, but the require_once directive should not be included in the `local` file. The full config block between line 25 and 116 is skipped if `wp-config-local.php` is found, but the local config shouldn't contain anything after 116 (unless you roll your own full config file)

https://github.com/pantheon-systems/WordPress/blob/default/wp-config.php#L116

greg-1-anderson:
Closing per ^^

PR #15, opened 2015-04-19
Re-working cache clearing to separate granular from site-wide

joshkoenig:
This is for review only. 

We will hopefully have this ironed out in time to merge as part of the 4.2 release.


joshkoenig:
This is now deprecated. It's all done here!

https://wordpress.org/plugins/pantheon-advanced-page-cache/

PR #16, opened 2015-06-02
Move WP_DEBUG to the non-local block

stueynet:
The issue is that I want to define WP_DEBUG in my wp-config-local.php file. In my local environment I turn WP_DEBUG on but in the live environment I want it off. The current setup has WP_DEBUG defined in both places which causes a warning when you set the wp-config-local.php WP_DEBUG to true. So the solution is to move the wp-config.php WP_DEBUG definition to the end of the "not using wp-config-local.php" block.

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,7 +20,7 @@ When the spin-up process is complete, you will be redirected to the site's dashb
 
 How about the WordPress database config screen? No need to worry about database connection information as that is taken care of in the background. The only step that you need to complete is the site information and the installation process will be complete.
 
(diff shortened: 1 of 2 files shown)


stueynet:
Looking closer it might make sense to have everything at the bottom moved up to the non local block. Since these would all be defined in the wp-config-local.php. Assuming of course the user is using a complete, standard wp-config.php file at wp-config-local.php.

jhabdas:
@stueynet I believe it's safe to say this isn't going to be merged anytime soon. Could you close this pull and open a related issue instead for discussion?

PR #19, opened 2015-08-06
Clean up WP coding standards for wp-config.php

westonruter:


westonruter:
I realize that if this is merged then it will cause problems for sites that have modified their `wp-config.php` based on this one. So perhaps the issues in the `wp-config.php` can't be rolled into the WordPress upstream and this PR should be closed.

PR #20, opened 2015-08-21
Ensure HTTPS gets used in WP_HOME and WP_SITEURL when enabled

westonruter:
Without this change, links generated by WP will use HTTP even when accessing the site over HTTPS. I noticed this specifically in the SSL admin, that images inserted into the post content were using HTTP srcs as opposed to HTTPS ones. This patch fixes the problem.


joshkoenig:
We need to fix this. Was re-filed as #54

joshkoenig:
Handled in #73

PR #21, opened 2015-09-04
gitignore directories likely to contain large binary objects

joemiller:
The Updraft and BackUpWordpress plugins are likely to generate very large binary objects inside the git-backed code directory of a website. These large objects should not be stored in git and will cause problems with workflows. This change will help keep these large files out of the git history.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,6 +5,9 @@ wp-content/uploads
 wp-content/blogs.dir/
 wp-content/upgrade/
 wp-content/backup-db/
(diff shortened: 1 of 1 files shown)


fujin:
LGTM! :+1:

joshkoenig:
This is rad. I will see if there's an upstream point release coming soon (4.3.1) and if not we can roll our our own update.

joshkoenig:
gonna move this into the next release which should be today

Issue #25, opened 2015-10-07
Pre-config WP_TEMP_DIR on Pantheon


joshkoenig:
We can prevent the unsafe use of global `/tmp` by supplying a `WP_TEMP_DIR` constant in our `wp-config.php`. It would point to `/srv/bindings/<binding-id>/tmp`.

This should be a one-liner in our area of the config file. Should be safe to roll into the next point release from upstream.

joshkoenig:
@greg-1-anderson @TeslaDethray heads up on this. We should get a PR ready for this and then roll it into the next upstream point release. It would be ideal for us to use this opportunity to get Sara leveled-up on helping to maintain the upstream.

danielbachhuber:
@joshkoenig There's two system-level implementations of this issue: https://github.com/pantheon-systems/WordPress/blob/master/wp-includes/functions.php#L1631-L1639

* `sys_temp_dir`, used by `sys_get_temp_dir()` is an ini setting available in PHP 5.5
* WordPress falls back to a `upload_tmp_dir` ini setting when `sys_get_temp_dir()` isn't available. 

To force the use of a safe temp directory, wouldn't it be preferable to do this at the system-level, instead of making it somewhat optional in wp-config?

If you want to do it within WordPress, #32 sets `WP_TEMP_DIR` in `wp-config.php` based on wp-config's path, as I don't see the `<binding-id>` exposed in an environment variable.

joshkoenig:
Fixed in #32

PR #32, opened 2015-10-28
Force the use of a safe temp directory when in a container

danielbachhuber:
See #25

### DIFF
--- a/wp-config.php
+++ b/wp-config.php
@@ -72,6 +72,12 @@ else:
     }
     // Don't show deprecations; useful under PHP 5.5
     error_reporting(E_ALL ^ E_DEPRECATED);
+    // Force the use of a safe temp directory when in a container
+    if ( 0 === stripos( dirname( __FILE__ ), '/srv/bindings/' ) ):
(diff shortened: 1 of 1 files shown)


wp-config.php
```diff
@@ -72,6 +72,12 @@
     }
     // Don't show deprecations; useful under PHP 5.5
     error_reporting(E_ALL ^ E_DEPRECATED);
+    // Force the use of a safe temp directory when in a container
+    if ( 0 === stripos( dirname( __FILE__ ), '/srv/bindings/' ) ):
```
joshkoenig: +1 for this change.

I think we should use the same mechanism for determining if we're "on Pantheon" we to elsewhere: looking for the presence of our signature constants. I'd check for `PANTHEON_BINDING`, which you can also use in place of `$pantheon_file_path_parts`, which will simplify this.

danielbachhuber: Ah. I didn't realize there was a `PANTHEON_BINDING` constant. Are they documented somewhere?

danielbachhuber: Updated in df2f3234928715f017a023a72ca89579505e4919

Issue #17, opened 2015-07-20
Allow for cache TTL of zero in cache plugin


timani:
In some instances such as development users dont want to have the minimum TTL of 600s.

danielbachhuber:
This will be coming in the 4.4 release.

PR #33, opened 2015-10-28
Abstract Pantheon Cache options read to a method

danielbachhuber:
This ensures the value is requested just-in-time, and adds a filter for
modifying the value on the fly.

See #17


danielbachhuber:
Actually, this will break backwards compat for anything using the `$options` or `$default_options` class variables.

danielbachhuber:
I'm also not sure where I got the idea I needed to add a filter for this.

PR #34, opened 2015-10-28
Only force 60 second ttl in `live` environment

danielbachhuber:
Other environments, including `test` and `dev`, are fine to use 0 second
ttls.

Also makes the behavior more consistent: if the ttl value is less than
60 seconds, then it should be forced to 60 seconds, not 600.

See #17

### DIFF
--- a/wp-content/mu-plugins/pantheon/pantheon-cache.php
+++ b/wp-content/mu-plugins/pantheon/pantheon-cache.php
@@ -201,8 +201,9 @@ class Pantheon_Cache {
 	 */
 	public function cache_add_headers() {
 		$ttl = absint( $this->options['default_ttl'] );
-		if ( $ttl < 60 )
-			$ttl = 600;
+		if ( $ttl < 60 && isset( $_ENV['PANTHEON_ENVIRONMENT'] ) && 'live' === $_ENV['PANTHEON_ENVIRONMENT'] ) {
+			$ttl = 60;
+		}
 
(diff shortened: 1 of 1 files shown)


danielbachhuber:
@joshkoenig #reviewmerge

ari-gold:
@danielbachhuber Happy to see this work to allow 0 seconds cache TTL for non-live environments!

@joshkoenig For Drupal sites, Launch Check on the "Status tab" recommends 15 minutes minimum for expiration of cached pages. Does it make sense to bump the default from 60 or 600 seconds to 900 seconds (15 minutes)?

joshkoenig:
The original code doesn't make any sense (if less than 60 set to 600???) but I don't know if the intention here to "force" a 60 second TTL is legit. If I want my live site to have a 0 ttl, why can't I?

danielbachhuber:
> If I want my live site to have a 0 ttl, why can't I?

Because it has the potential to cause unreasonable resource usage in production?

danielbachhuber:
I think being able to set `ttl=0` for `live` can be a later enhancement separate of this PR.

joshkoenig:
Ok. We generally hold releases here until there's something upstream to bundle them with. Do you want to start a `4.4` branch and change the pull target to that? That'll let us merge changes we plan on rolling out with the next WP core update.

danielbachhuber:
> . Do you want to start a 4.4 branch and change the pull target to that?

#37

PR #37, opened 2015-10-28
Only force 60 second ttl in `live` environment (release-4.4)

danielbachhuber:
Other environments, including `test` and `dev`, are fine to use 0 second
ttls.

Also makes the behavior more consistent: if the ttl value is less than
60 seconds, then it should be forced to 60 seconds, not 600.

See #17 
Conversation in #34


joshkoenig:
Seems legit

PR #45, opened 2015-10-29
Register `wp pantheon cache clear` for purging cache based on path

### LINKED ISSUE #27: WP-CLI interface for cache clearing
The interface for "clear all" works fine, but we should also have a CLI interface to purge individual urls. 

This is definitely similar to #28, but I believe we should be able to define the interface for this even if it uses the old BAN method internally.

I would imagine the steps to be something like:

1. Distinguish the code paths for "clear all" and "clear this array of paths" in `pantheon-cache.php`
2. Make sure that the general cache clear command uses clear all, and that there's a dedicated CLI for clearing a specific set of up to 10 paths.
3. Make sure that the specific path-clearing operation uses the preferable PURGE api instead of the old BAN api.

danielbachhuber:
```
salty-wordpress ➜  pantheon.dev  wp pantheon cache clear
Error: Please provide one or more URL paths to purge, or use the --all
flag.
```

```
salty-wordpress ➜  pantheon.dev  wp pantheon cache clear /foo/ /bar/
Success: Cache purged for 2 URL paths.
```

```
salty-wordpress ➜  pantheon.dev  wp pantheon cache clear --all
Success: Cache purged for all URL paths.
```

Fixes #27

### DIFF
--- a/wp-content/mu-plugins/pantheon.php
+++ b/wp-content/mu-plugins/pantheon.php
@@ -13,4 +13,8 @@ if ( isset( $_ENV['PANTHEON_ENVIRONMENT'] ) ) :
 require_once( 'pantheon/pantheon-cache.php' );
 require_once( 'pantheon/pantheon-updates.php' );
 
+if ( defined( 'WP_CLI' ) && WP_CLI ) {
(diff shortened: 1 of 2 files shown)


danielbachhuber:
@joshkoenig #reviewmerge

PR #46, opened 2015-10-30
Display a more helpful message if WP Redis has errors on Pantheon

danielbachhuber:
Depends on https://github.com/pantheon-systems/wp-redis/pull/18
See https://github.com/pantheon-systems/wp-redis/issues/9#issuecomment-152637365


danielbachhuber:
Ends up looking like this:

image

PR #48, opened 2015-11-17
Ignore directories likely to contain large binary objects

joshkoenig:
The Updraft and BackUpWordpress plugins are likely to generate very large binary objects inside the git-backed code directory of a website. These large objects should not be stored in git and will cause problems with workflows. This change will help keep these large files out of the git history.


joshkoenig:
Picking up for #21 to get into the next release to master.

PR #52, opened 2015-12-16
Register pantheon-messages to direct users in git mode to the dashboard

bmackinney:
## While in git mode

Currently, users who attempt to update a plugin or theme see this:

image

This pull request replaces that with 
image

When users attempt to install a new plugin or theme they see: 
image

This pull request replaces that with:
image

I'm sure there is a simple reason the message is duplicated on the installation screen.

I started this Pull Request to run on live and direct users back to Dev, but that is no longer necessary since the addition of `DISALLOW_FILE_MODS`

### DIFF
--- a/wp-config.php
+++ b/wp-config.php
@@ -77,11 +77,6 @@ else:
         define( 'WP_TEMP_DIR', sprintf( '/srv/bindings/%s/tmp', PANTHEON_BINDING ) );
     endif;
 
-    // FS writes aren't permitted in test or live, so we should let WordPress know to disable relevant UI
-    if ( in_array( $_ENV['PANTHEON_ENVIRONMENT'], array( 'test', 'live' ) ) && ! defined( 'DISALLOW_FILE_MODS' ) ) :
-      define( 'DISALLOW_FILE_MODS', true );
-    endif;
-
   else:
     /**
      * This block will be executed if you have NO wp-config-local.php and you


joshkoenig:
@bmackinney this is a good change, but we should keep the `DISALLOW_FILE_MODS` for test and live. Those can never be toggled into SFTP mode. ;)

wp-content/mu-plugins/pantheon/pantheon-messages.php
```diff
@@ -0,0 +1,20 @@
...
+if (defined('PANTHEON_ENVIRONMENT')) {
+  // Only when filesystem credentials are requested when in development environments in Git mode
+  if(has_filter('request_filesystem_credentials')){
+  apply_filter('request_filesystem_credentials');
+}
+    // Stop users from trying to install plugins in Git mode
+    add_filter( 'request_filesystem_credentials', 'pantheon_modify_ftp_request', 8 );
+        function pantheon_modify_ftp_request () {
+          $PANTHEON_DASHBOARD = 'https://dashboard.pantheon.io/sites/'.$_ENV['PANTHEON_SITE'].'#'.$_ENV['PANTHEON_ENVIRONMENT'].'/code';
+    echo 'Plugins and Themes cannot be installed while this environment is in Git mode. Please return to the <a href="'.$PANTHEON_DASHBOARD.'">Pantheon Site Dashboard</a> and switch the connection mode to SFTP.';
```
ataylorme: We should use proper escaping here to maintain WordPress coding standards

wp-config.php
```diff
@@ -77,11 +77,6 @@
         define( 'WP_TEMP_DIR', sprintf( '/srv/bindings/%s/tmp', PANTHEON_BINDING ) );
     endif;
 
-    // FS writes aren't permitted in test or live, so we should let WordPress know to disable relevant UI
-    if ( in_array( $_ENV['PANTHEON_ENVIRONMENT'], array( 'test', 'live' ) ) && ! defined( 'DISALLOW_FILE_MODS' ) ) :
-      define( 'DISALLOW_FILE_MODS', true );
-    endif;
```
westonruter: I don't think this should be removed. WordPress should still be blocked from modifying the filesystem on non-dev environments.

wp-content/mu-plugins/pantheon/pantheon-messages.php
westonruter: This whole file should be run through the `WordPress-Core` and `WordPress-Extra` standards with PHP_CodeSniffer.

bmackinney:
IIRC there was a core bug related to this PR. Good luck! closing because, you know.

PR #56, opened 2016-01-08
Exclude .htaccess files since Pantheon uses nginx webservers and does…

grok:
Exclude .htaccess files since Pantheon uses nginx webservers and does not recognize or parse Apache's directory-level configuration files.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,12 +13,12 @@ wp-content/advanced-cache.php
 wp-content/wp-cache-config.php
 sitemap.xml
 sitemap.xml.gz
+.htaccess
 *.log
 
 # @TODO writable paths
 wp-content/cache/
 wp-content/backups/
-
 
 # Packages #
 ############


greg-1-anderson:
I don't think this is necessarily desirable. If someone exports a Pantheon site and serves it with Apache, it'd be better to still have the .htaccess files present. In any event, having these present on Pantheon has a minimal impact and no adverse effects.

grok:
I agree, that the impact is minimal.

The adverse effect is that some codebases contain files that are not and will not be in play.
This can cause some situations where someone has an issue with permalinks, or redirects and may not be familiar with Pantheon and thinks that because the file exists, that it is being utilized.

From a house cleaning perspective, it's bloat.

From a migration perspective, WordPress actually generates this file on it's own when asked to.

There is a remote usecase where keeping it makes sense, which is that the .hattaccess is highly customized and someone moved from Jim Bob's hosting to Pantheon with the expectation of leaving to another hosting platform later.

OOOORRRR

Their development outside the Pantheon platform involves Apache. Which I think should be discouraged as NGINX development and Apache development are very much not apples to apples.

On a scale of 1 to 10 on how important this suggestion is... 1 being analogous to "remove this whitespace" and 10 being "this will improve everyones website by 50%" ... this is a 1.3? 😃

greg-1-anderson:
Pantheon has a policy of not modifying the upstream whenever possible. This trumps the other concerns here.

grok:
@greg-1-anderson ... of all the reasons you could have POSSIBLY given... that was actually pretty solid.

Thanks for the time 👍

Issue #57, opened 2016-01-18
protocol should be set dynamically for WP_HOME and WP_SITEURL


gmcinnes:
I ran into a problem when upgrading to WP 4.4.  WP 4.4. introduces responsive images, and creates image tags like this:

```
<img width="320" height="225" src="https://www.marsdd.com/wp-content/uploads/2015/05/governance-collections.png" class="attachment-board-thumb size-board-thumb wp-post-image" alt="governance-collections" srcset="http://www.marsdd.com/wp-content/uploads/2015/05/governance-collections-300x211.png 300w, http://www.marsdd.com/wp-content/uploads/2015/05/governance-collections.png 320w" sizes="(max-width: 320px) 100vw, 320px">
```

In my case the site was being served from **https**, but the WP_SITEURL and WP_HOME variables in `wp-config.php` are hard-coded to point to **http**.

The `srcset` attributes generated in the tag above are generated from the WP_SITEURL setting, they use an _http_ url, not an _https_ one.  Both Chrome and Safari refuse to load an image in that case, causing site breakage.

Perhaps the protocol parts of WP_SITEURL and WP_HOME should also be set dynamically depending on the request as well?

gmcinnes:
I should add  that this issue was only happening on my live environment.  On my multidev environments WP_HOME and WP_SITEURL are being set correctly based on the protocol the site is accessed with (based on viewing the admin / settings page in both http and https).

 I don't see anything in the code at the moment that would indicate why that's happening though.

ahebrank:
I'm curious about this as well.  Can Pantheon comment on why the protocol is hardcoded to http://?

joshkoenig:
@ahebrank there's not a good reason. I believe we can adjust our code to automatically do the right thign based on `HTTP_X_FORWARDED_PROTO`. Checking that now.

dylanjameswagner:
This issue appears to be resolved in #73, this issue can be closed.

nullvariable:
@greg-1-anderson per the previous comment, can you close this issue? :pray: :bowing_man:

PR #58, opened 2016-01-18
Remove hard-coded protocol from WP_SITEURL and WP_HOME

gmcinnes:
WP_SITEURL and WP_HOME were hard-coded to use `http` as the protocol.  Maybe they should set the protocol based on the site request, in the same way as they do with the HTTP_HOST.

This may need tweaking to work in the Pantheon environment. YMMV.

Closes pantheon-systems/Wordpress#57

### DIFF
--- a/wp-config.php
+++ b/wp-config.php
@@ -66,10 +66,12 @@ else:
     /**#@-*/
 
     /** A couple extra tweaks to help things run well on Pantheon. **/
+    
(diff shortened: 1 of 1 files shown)


joshkoenig:
We need to fix this for sure, but I'm not sold on the method just yet.

joshkoenig:
Handled in #73

PR #69, opened 2016-03-14
Flush the WordPress object cache when flushing an entire site

danielbachhuber:
See https://github.com/pantheon-systems/wp-redis/issues/72


danielbachhuber:
@joshkoenig #reviewmerge

joshkoenig:
This looks good. I'm wondering more and more about whether that `wp_redirect` really belongs there, but we can address this separately.

I don't think we need to make a special release for this, but we should roll it into the next upstream update.

danielbachhuber:
> I don't think we need to make a special release for this, but we should roll it into the next upstream update.

@joshkoenig Is a PR against master fine for this, or is there something else you'd like me to do?

joshkoenig:
This is fine. We just need to make sure we remember (paging @greg-1-anderson...) to merge this in with the next upstream release.

greg-1-anderson:
Fixed in #72.

PR #72, opened 2016-04-12
Update to WordPress 4.5

greg-1-anderson:
For more information, see https://wordpress.org/news/2016/04/coleman/

PR #73, opened 2016-04-12
Use our USER_AGENT_HTTPS header to set the default scheme

### LINKED ISSUE #57: protocol should be set dynamically for WP_HOME and WP_SITEURL

joshkoenig:
Closes #57 #58 
Fixes #20

### DIFF
--- a/wp-config.php
+++ b/wp-config.php
@@ -67,8 +67,16 @@ else:
 
     /** A couple extra tweaks to help things run well on Pantheon. **/
     if (isset($_SERVER['HTTP_HOST'])) {
-      define('WP_HOME', 'http://' . $_SERVER['HTTP_HOST']);
-      define('WP_SITEURL', 'http://' . $_SERVER['HTTP_HOST']);
+        // HTTP is still the default scheme for now. 
+        $scheme = 'http';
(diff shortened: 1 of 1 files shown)


greg-1-anderson:
Fixed in #72

PR #75, opened 2016-04-29
Include wp-redis object-cache.php if present.

vincenzo:
This should avoid developers to add/remove the object-cache.php "symlink" when installing/removing `wp-redis`.

### DIFF
--- a/wp-content/object-cache.php
+++ b/wp-content/object-cache.php
@@ -1,0 +1,7 @@
+<?php
+// Path to the wp-redis object-cache.php file.
+$wpredis_file = WP_CONTENT_DIR . '/plugins/wp-redis/object-cache.php';
+// Requires the file only if present (i.e. if the wp-redis plugin is present).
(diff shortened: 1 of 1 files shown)


greg-1-anderson:
@joshkoenig @danielbachhuber Thoughts?

danielbachhuber:
@greg-1-anderson This will cause a merge conflict with any `object-cache.php` already tracked in Git.

vincenzo:
? 
Could you give an example so it's clearer for me?

danielbachhuber:
> Could you give an example so it's clearer for me?

When a new Pantheon site is created, it's done so with a Git clone of this upstream repo. This Git clone is how new WordPress versions are deployed to Pantheon sites.

If that new site adds an `wp-content/object-cache.php` file, then it's added to the Git version history. As such, if we add a `wp-content/object-cache.php` in this repo, then this repository's newer version would conflict with the Pantheon site's older version when the site merges in the latest changes from this repo.

PR #76, opened 2016-05-03
Update .gitignore

MikeNGarrett:
I recommend removing `Icon?` from the gitignore as it matches files and directories in a case-insensitive way named "icons" or "icon1" etc

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -42,7 +42,6 @@ wp-content/backups/
 ######################
 .DS_Store*
 ehthumbs.db
-Icon?
 Thumbs.db
 ._*
 


greg-1-anderson:
@joshkoenig What pattern is best here? Drops-6 is using `Icon?`, whereas drops-7 and drops-8 both use `Icon`.  What file patterns are actually intended to be excluded by this rule?  Would `Icon\?` be better, or should we just take this out as recommended by this PR?

joshkoenig:
Case-sensitivity is up to your operating system/filesystem settings. I believe the intent here is to exclude cruft that can show up from MacOS specifically, and the `?` was there to account for a few different cases.

@MikeNGarrett  - Is the challenge that you have a directory called `icons` you want to add at the root level?

ataylorme:
@MikeNGarrett rather than removing `Icon?` from `.gitignore` can you update the PR so that line is `/Icon?`? 

According to [their docs page](https://git-scm.com/docs/gitignore#_pattern_format) a slash in `.gitignore` will restrict the ignore to the top-level rather than being recursive.

MikeNGarrett:
@ataylorme Happy to do that, but isn't the point of this rule to eliminate files that show up throughout the repo, not just in the root?

ataylorme:
@MikeNGarrett sorry I might have misunderstood the issue. What is the directory name of the specific item you've had issues with? And where is it located (e.g. the theme)?

After researching a bit more it looks like the file we want to ignore is literally named `Icon?` but the `?` actually stands for a double carriage return. We can do something like `printf "Icon\r\r" >> .gitignore` to add it correctly or negate icons directories with `!icons` following the current rule.

MikeNGarrett:
@ataylorme I had problems with directories inside themes and plugins that were ignored, eg /wp-content/themes/theme-name/images/icons/*

I assume `Icon` files are system files on some OS that should be ignored, but I'm not sure what the original files are or from what system. I don't know if there's a solution that will block these types of files and allow directories or files that use "icon" in their file path or name.

ataylorme:
@MikeNGarrett Yes, it is a file created by OS X when you add a custom icon to a directory in Finder.

> We can do something like `printf "Icon\r\r" >> .gitignore` to add it correctly or negate icons directories with `!icons` following the current rule.

@joshkoenig and @greg-1-anderson preferences on those options? Other ideas?

greg-1-anderson:
I am opposed to adding invisible characters to the .gitignore file. It would be too easy to remove these by accident in future edits.

jhabdas:
This is growing stale. Is there still discussion to be had here, or can this be closed?

MikeNGarrett:
Ok, I think I know what the root of the problem is now. 

Your gitignore converts the end of line characters to a `?` leaving `Icon?` and immediately goes to the next rule. The Mac OS gitignore requires 2 end-of-line characters, specifically `\r\r` to match correctly. 

Reference: https://github.com/github/gitignore/blob/master/Global/macOS.gitignore

dzmarc:
Reigniting this thread because we recently ran into an issue that consumed quite a bit of time for several individuals including plugin devs, our internal devs and some Pantheon support. A few of our plugins had directory locations with the word "icon" in it that were being ignored, causing some functionality to be missing or throwing errors on the Pantheon environment. 

Since this is Platform related, we would like to see some sort of fix or suggested workaround that would not require us to either tediously dig through all of our vendor plugins to see if there is something we need to manually force add or edit the upstream gitignore and conflict with any upstream updates.

dylanjameswagner:
I have had good luck with adding an exception for directories named `icons`. It solves most of my issues, however there are still some cases that this may not cover.

```
Icon?
## @link https://github.com/pantheon-systems/WordPress/issues/66
## @link https://github.com/pantheon-systems/WordPress/pull/76
## @link http://stackoverflow.com/questions/17556250/how-to-ignore-icon-in-git
!icons/
```

dzmarc:
There are several ways listed here to work around it. The only issue is that, pulling down the latest upstream will override any workarounds we apply.

greg-1-anderson:
Yeah, this should have been merged years ago (or at least when the drops-8 PR was merged last July)

MikeNGarrett:
Yay! Finally! 🎉

PR #92, opened 2016-12-22
Update .gitignore

MikeNGarrett:
The `?` at the end of Icon matches any directory using this word "icon*" or "Icon*" which is bad news when you use plugins or themes with icon files or directories.


MikeNGarrett:
I forgot I already submitted this. Closing.

PR #102, opened 2017-03-21
Correct Salt Documentation

### LINKED ISSUE #86: wp-config mentions changing unique salts from dashboard
Issue raised in a support chat with @mattpaz.

Our wp-config.php mentions

> Pantheon sets these values for you also. If you want to shuffle them you can do so via your dashboard.

here: https://github.com/pantheon-systems/WordPress/blob/master/wp-config.php#L53

We don't have any good documentation on how to do this, is this still available?

rachelwhitton:
Re: DSI-67 
Closes #86 

The feature to shuffle salts via the Dashboard has not been implemented/delivered. Suggest users contact support to shuffle salts instead.

### DIFF
--- a/wp-config.php
+++ b/wp-config.php
@@ -51,7 +51,7 @@ else:
      * You can change these at any point in time to invalidate all existing cookies. This will force all users to have to log in again.
      *
      * Pantheon sets these values for you also. If you want to shuffle them you
-     * can do so via your dashboard.
+     * must contact support: https://pantheon.io/docs/getting-support 
      *
      * @since 2.6.0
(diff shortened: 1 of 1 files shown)


JJJ:
This is an easy one, and good to have correct.

While not super common, if you do ever need to update these, it's very typical someone would just go get new ones from the [wporg](https://api.wordpress.org/secret-key/1.1/salt/) generator.

If that's not the case here, it should be documented as such.

greg-1-anderson:
Deployed with WordPress 4.8

PR #126, opened 2017-07-05
Replace WordPress update nag with our own

ataylorme:
Edit: **The process for updating WordPress on Pantheon is different. We need to tell people that (or remind them).**

I refactored our mu-plugin. Previously it stopped WordPress from running version checks.

I allowed WordPress to run version checks then use the result to determine if we should display our own nag message, removing the default.

The previous logic also only stopped WordPress updates on test and live. The new nag message applies to all environments.

It looks like this:
pantheon-update-nag

### DIFF
--- a/wp-content/mu-plugins/pantheon/pantheon-updates.php
+++ b/wp-content/mu-plugins/pantheon/pantheon-updates.php
@@ -34,7 +34,7 @@ function _pantheon_wordpress_update_available() {
 	include( ABSPATH . WPINC . '/version.php' );
 	
 	// Return true if our version is not the latest
-	return $latest_wp_version > $wp_version;
+	return version_compare( $latest_wp_version, $wp_version, '>' );
 
 }
 
--- a/wp-content/mu-plugins/pantheon/pantheon-updates.php
+++ b/wp-content/mu-plugins/pantheon/pantheon-updates.php
@@ -1,5 +1,5 @@
 <?php
-// Disable WordPress ato updates
+// Disable WordPress auto updates
 if( ! defined('WP_AUTO_UPDATE_CORE')) {
 	define( 'WP_AUTO_UPDATE_CORE', false );
 }
@@ -61,6 +61,15 @@ function _pantheon_register_upstream_update_notice(){
 	if( isset( $_ENV['PANTHEON_ENVIRONMENT'] ) && _pantheon_wordpress_update_available() ){
 		add_action( 'admin_notices', '_pantheon_upstream_update_notice' );
 	}
+}
+
+function _pantheon_disable_wp_updates() {
+	include ABSPATH . WPINC . '/version.php';
+	return (object) array(
+		'updates' => array(),
+		'version_checked' => $wp_version,
+		'last_checked' => time(),
+	);
 }
 
 // Only in Test and Live Environments...
--- a/wp-content/mu-plugins/pantheon/pantheon-updates.php
+++ b/wp-content/mu-plugins/pantheon/pantheon-updates.php
@@ -46,9 +46,9 @@ function _pantheon_upstream_update_notice() {
 	$latest_wp_version = _pantheon_get_latest_wordpress_version();
     ?>
     <div class="update-nag">
-		<h3>
-			"WordPress <?php echo $latest_wp_version; ?> is available! Please update to WordPress <?php echo $latest_wp_version; ?> from <a href="https://dashboard.pantheon.io/sites/<?php echo $_ENV['PANTHEON_SITE']; ?>">your Pantheon dashboard</a>.
-		</h3>
+		<p style="font-size: 14px; font-weight: bold; margin: 0 0 0.5em 0;">
+			WordPress <?php echo $latest_wp_version; ?> is available! Please update from <a href="https://dashboard.pantheon.io/sites/<?php echo $_ENV['PANTHEON_SITE']; ?>">your Pantheon dashboard</a>.
+		</p>
 		For details on applying updates, see <a href="https://pantheon.io/docs/upstream-updates/" target="_blank">the Applying Upstream Updates documentation</a>. <br />
 		If you need help, open a support chat on Pantheon.
 	</div>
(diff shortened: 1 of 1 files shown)


ataylorme:
@greg-1-anderson first pass at this. We can easily update the `_pantheon_get_latest_wordpress_version` function later to check our upstream rather than WordPress core versions if that becomes an option in the future.

wp-content/mu-plugins/pantheon/pantheon-updates.php
```diff
@@ -1,35 +1,76 @@
...
-	// Disable Plugin Updates
-	//
-	add_action('admin_menu','_pantheon_hide_admin_notices');
-	function _pantheon_hide_admin_notices() {
-		remove_action( 'admin_notices', 'update_nag', 3 );
+// Compare the current WordPress version to the latest available
+function _pantheon_wordpress_update_available() {
+	$latest_wp_version = _pantheon_get_latest_wordpress_version();
+
+	if( null === latest_wp_version ){
```
westonruter: Missing `$` here.

wp-content/mu-plugins/pantheon/pantheon-updates.php
```diff
@@ -1,35 +1,76 @@
...
+		return false;
+	}
+
+	// include an unmodified $wp_version
+	include( ABSPATH . WPINC . '/version.php' );
+	
+	// Return true if our version is not the latest
+	return $latest_wp_version > $wp_version;
```
westonruter: Need to use `version_compare()` here since `version_compare('3.11', '3.8', '>') !== ( '3.11' > '3.8' )`.

wp-content/mu-plugins/pantheon/pantheon-updates.php
```diff
@@ -1,35 +1,76 @@
...
+function _pantheon_upstream_update_notice() {
+	// include an unmodified $wp_version
+	include( ABSPATH . WPINC . '/version.php' );
+
+	$latest_wp_version = _pantheon_get_latest_wordpress_version();
+    ?>
+    <div class="update-nag">
+	<h3>Please check <a href="https://dashboard.pantheon.io/sites/<?php echo $_ENV['PANTHEON_SITE']; ?>"> to see if a WordPress update is available on the platform.</h3>
+		Your WordPress version of <?php echo $wp_version; ?> is out of date (the current version is <?php echo $latest_wp_version; ?>)<br />
+		If you need help, please see <a href="https://pantheon.io/docs/upstream-updates/">the documentation</a> for details or open a support chat.
```
westonruter: Is translatability a concern?

ataylorme: I'm not sure. I can certainly add translation functions and generate a `.pot` file for the plugin but don't know if we have plans to actually translate anything (currently the Pantheon dashboard in English only).

Maybe @joshkoenig can weigh in.

westonruter: I suppose the plugin isn't on WordPress.org, huh. I was thinking this was the https://wordpress.org/plugins/search/pantheon-hud/ but I see it isn't. If it were on the directory, then the translations could be crowd sourced via translate.wordpress.org

ataylorme: Nope, it's just in our custom upstream and I doubt anyone is running it outside of Pantheon outside of a few folks who have migrated off the platform but they wouldn't see this update anyway

ataylorme:
@greg-1-anderson mind reviewing this?

greg-1-anderson:
+1 LGTM

If you want to get a list of all SemVer tags on the Pantheon WordPress repository:
```
git ls-remote --tags git@github.com:pantheon-systems/WordPress.git  | sed -e 's|refs/tags/||' | cut -f2 | grep '^[0-9]\+\.[0-9]\+[0-9\.]*$' | sort -r -n
```
It would probably be better to php-ize that, rather than exec'ing the shell. Not sure if there's an easy way for us to get our upstream. It's available in the metadata.json file; however, that file is not readable by the app server.

It's fine by me to put off improvements to version checking until laer.

ari-gold:
@ataylorme  @greg-1-anderson This is great! I'd like to take a pass at refining the copy before releasing. /cc if you have suggestions: @informanddelight  @kimby77

joshkoenig:
Seems legit. +1 for getting sign-off from docs/design on verbage.

With that, I think this could roll out with the next core WP update.

kimby77:
@ataylorme 
Am I understanding this correctly, that this message is to encourage people to apply updates from the pantheon dashboard vs. the WP admin?  If so, we can be more authoritative about it in the language so it comes across as less of a suggestion and more of an instruction.  That said, are there cases to account for if...

1) Users **shouldn't** be applying the update in Pantheon vs WP admin? We don't want to direct them to do the wrong thing.
2) If the WP update isn't available? The current heading reads "Please check the Pantheon dashboard to see if a WordPress update is available on the platform." The "if" reads a little ambiguous - maybe the update is there, maybe it's not, and what am I supposed to do if it's not?

greg-1-anderson:
This PR is an improvement over the default action, which instructs the user to update from the WP Admin panel, which is always wrong.

This PR does not account for:

- The delay between the availability of the WordPress update and the availability of the same update on Pantheon's WordPress upstream (usually same-day, but there is a delay).
- The possibility that the site might be managed by Composer (in which case the site should be updated by Composer, not by the dashboard).
- The possibility that the site might be using a custom upstream (in which case there will be an additional delay before someone makes the update available for the custom upstream).

This is the reason for the somewhat vague "if".

ari-gold:
How does the stock/vanilla WP nag read?

ataylorme:
> How does the stock/vanilla WP nag read?

It reads: "WordPress <version> is available! Please update now." with "update now" linked to the WordPress update functionality.

greg-1-anderson:
screen shot 2017-07-18 at 2 39 34 pm

kimby77:
Thanks for the explanation @greg-1-anderson!

I suggest something like:

"WordPress 4.8 is available! Please update from your [Pantheon dashboard](url). 

For details on applying updates, see the [Applying Upstream Updates](https://pantheon.io/docs/upstream-updates/) documentation. If you need help, open a support chat on Pantheon."

I left out info on the update delay in the interest of brevity here, so we might keep tabs on whether we get a lot of questions in about that and then we could update this message and/or the linked documentation to communicate that.

ataylorme:
@kimby77 updated, here ya go:

pantheon-wordpress-update-nag-updated

kimby77:
@ataylorme a nitpick is to have the link set up like so: "see the [link]Applying Upstream Updates[/link] documentation" so that the link refers to only name of the document (ie. the destination) and is not too long.

Otherwise looks great to me!

ataylorme:
> have the link set up like so: "see the [link]Applying Upstream Updates[/link] documentation", so that the link refers to only name of the document (ie. the destination) and is not too long.

Updated

kimby77:
👍

Issue #135, opened 2017-09-25
Ignore differences to $wp_version revision number when wordpress-develop (src) is deployed


westonruter:
I have an environment where I have `SCRIPT_DEBUG` enabled and am pushing out the files directly from wordpress-develop where the `$wp_version` is `4.9-alpha-40870-src`. Even though I have the latest from `trunk`, in the dashboard I see this message:

![image](https://user-images.githubusercontent.com/134745/30825168-e2998bcc-a1e6-11e7-9d11-7df4502bc95d.png)

I think it is because the nightly build bumps the changeset rev number in `$wp_version` whereas each commit to `wordpress-develop` does not (it only gets bumped with each beta/rc/final).

So I think that if `-src` appears in `$wp_version` that the revision number should be ignored from consideration when determining if there are new changes to update to.

ataylorme:
@westonruter can you test #136?

westonruter:
@ataylorme tested, feedback provided.

PR #136, opened 2017-09-25
Don't show WordPress update warning if the WordPress version contains src

ataylorme:
Addresses #135

### DIFF
--- a/wp-content/mu-plugins/pantheon/pantheon-updates.php
+++ b/wp-content/mu-plugins/pantheon/pantheon-updates.php
@@ -19,8 +19,9 @@ function _pantheon_get_latest_wordpress_version() {
 		return null;
 	}
 
-	// We're only looking for stable releases so return null for trunk or beta version of WordPress
-	if( false === stripos($core_updates[0]->current, 'beta') || false === stripos($core_updates[0]->current, 'src') ){
+	// We're only looking for stable releases so return null for unstable versions of WordPress, such as trunk or beta
+	// Stable releases only contain numbers and periods, e.g. 4.9
+	if( 0 === preg_match('/^[\d\.]+$/', $core_updates[0]->current) ){
 		return null;
 	}
 
--- a/wp-content/mu-plugins/pantheon/pantheon-updates.php
+++ b/wp-content/mu-plugins/pantheon/pantheon-updates.php
@@ -15,7 +15,7 @@ function _pantheon_hide_update_nag() {
 function _pantheon_get_latest_wordpress_version() {
 	$core_updates = get_core_updates( array('dismissed' => false) );
 
-	if( ! is_array($core_updates) || empty($core_updates) || ! property_exists($core_updates[0], 'current' ) ){
+	if( ! is_array($core_updates) || empty($core_updates) || ! property_exists($core_updates[0], 'current' ) || false === stripos($core_updates[0]->current, 'src') ){
 		return null;
 	}
 
--- a/wp-content/mu-plugins/pantheon/pantheon-updates.php
+++ b/wp-content/mu-plugins/pantheon/pantheon-updates.php
@@ -15,7 +15,12 @@ function _pantheon_hide_update_nag() {
 function _pantheon_get_latest_wordpress_version() {
 	$core_updates = get_core_updates( array('dismissed' => false) );
 
-	if( ! is_array($core_updates) || empty($core_updates) || ! property_exists($core_updates[0], 'current' ) || false === stripos($core_updates[0]->current, 'src') ){
+	if( ! is_array($core_updates) || empty($core_updates) || ! property_exists($core_updates[0], 'current' ) ){
+		return null;
+	}
+
+	// We're only looking for stable releases so return null for trunk or beta version of WordPress
+	if( false === stripos($core_updates[0]->current, 'beta') || false === stripos($core_updates[0]->current, 'src') ){
 		return null;
 	}
 


westonruter:
@ataylorme While this works, it will have the unintended side effect of not prompting for updates when they are available. For example, if I have deployed `4.9-alpha-40870-src` but then a nightly build of `4.9-beta1-45000` is available, then I _should_ still get prompted to update. *But* it should not update using the nightly build because then that will blow away the `-src`. So in that case, a different notification should be displayed. 

In either case, however, I don't think that linking to the Pantheon dashboard is the right place to go in the case of nightly/trunk deployments. The upstream from Pantheon won't have any updates. If the beta tester plugin is enabled, then I think this should have instructions on turning on SFTP mode and then linking to the core update screen. Otherwise, if it is a `trunk` wordpress-develop (`src`) version, then the instructions should just direct the user to update the codebase from their own machine.

ataylorme:
>  If the beta tester plugin is enabled, then I think this should have instructions on turning on SFTP mode and then linking to the core update screen. Otherwise, if it is a trunk wordpress-develop (src) version, then the instructions should just direct the user to update the codebase from their own machine.

I think this might be out of scope. Let me know if you disagree but I would assume anyone running the beta tester plugin or trunk wordpress-develop knows how to update, as they did so to get to that state, and we don't need to provide instructions.

The notification that's provided now is only intended to display when a Pantheon upstream update for WordPress core to a higher version is available.

westonruter:
OK, sounds good to me.

wp-content/mu-plugins/pantheon/pantheon-updates.php
```diff
@@ -19,6 +19,12 @@ function _pantheon_get_latest_wordpress_version() {
 		return null;
 	}
 
+	// We're only looking for stable releases so return null for unstable versions of WordPress, such as trunk or beta
+	// Stable releases only contain numbers and periods, e.g. 4.9
+	if( 0 === preg_match('/^[\d\.]+$/', $core_updates[0]->current) ){
```
ataylorme: @westonruter how's this? Anything that doesn't contain a number or period will be marked as an unstable version and won't trigger the Pantheon update nag. I tested with standard versions like `4.8.2` and `4.9`.

westonruter: That works, though you could also just check to see if the version contains `-`.

danielbachhuber: If I've installed `4.9-alpha-40870-src` and `4.9` comes out, will I receive an update nag?

ataylorme: No, returning `null` here based on the current version not being a stable release will disable all update nags.

danielbachhuber: > No, returning `null` here based on the current version not being a stable release will disable all update nags.

Hm. Personally, I'd expect to see an update nag if there's a stable release that's more recent than the development version I'm running. I often forget to keep development versions up to date :/

We could write a version comparison function that works with custom WP versions, and do so in a separate repo with tests, etc.

Frankly, WP-CLI would benefit from this too because Weston has just highlighted a couple of inadequacies to my existing helper function: https://github.com/wp-cli/wp-cli/blob/master/php/utils.php#L246

danielbachhuber: https://github.com/wp-cli/wp-cli/issues/4380

westonruter: @danielbachhuber I think the comparison logic would be something like this:

```php
/**
 * Compare two WordPress version numbers which look like '4.9-alpha-40870-src'.
 *
 * @param string $ver_a Version A.
 * @param string $ver_b Version B.
 * @return int Comparison.
 */
$compare = function( $ver_a, $ver_b ) {
	$ver_a_parts = explode( '-', $ver_a );
	$ver_b_parts = explode( '-', $ver_b );

	// Compare semantic versions.
	$semver_a = array_shift( $ver_a_parts );
	$semver_b = array_shift( $ver_b_parts );
	$comparison = version_compare( $semver_a, $semver_b );
	if ( 0 !== $comparison ) {
		return $comparison;
	}

	// Compare pre-release identifiers 'alpha', 'beta', 'rc'.
	$prerelease_a = array_shift( $ver_a_parts );
	$prerelease_b = array_shift( $ver_b_parts );
	$comparison = strcmp( $prerelease_a, $prerelease_b );
	if ( 0 !== $comparison ) {
		return $comparison;
	}

	// If either version is a src (but not both) then consider them equal.
	$src_a = 'src' === array_shift( $ver_a_parts );
	$src_b = 'src' === array_shift( $ver_b_parts );
	if ( $src_a xor $src_b ) {
		return 0;
	}

	// Compare revision numers.
	$rev_a = intval( array_shift( $ver_a_parts ) );
	$rev_b = intval( array_shift( $ver_b_parts ) );
	return $rev_b - $rev_a;
};
```

ataylorme:
> I'd expect to see an update nag if there's a stable release that's more recent than the development version I'm running. I often forget to keep development versions up to date :/

Any objection to merging what we have now and then swapping out for a more advanced check later?

greg-1-anderson:
I think a little too much nag is better than not enough nag, if we have to chose a less-than-perfect solution.

danielbachhuber:
Given the level of effort in making changes to the upstream, I'd think it's better to get it right the first time.

If no one steps forth on the WP-CLI issue in the next couple of days, I can own this early next week.

danielbachhuber:
Closing in favor of #138

PR #138, opened 2017-10-03
Strip '-src' from `$wp_version` for more accurate version comparison

### LINKED ISSUE #135: Ignore differences to $wp_version revision number when wordpress-develop (src) is deployed

danielbachhuber:
Fixes #135
See original conversion in #136 and discovery with tests in https://github.com/wp-cli/wp-cli/pull/4395

### DIFF
--- a/wp-content/mu-plugins/pantheon/pantheon-updates.php
+++ b/wp-content/mu-plugins/pantheon/pantheon-updates.php
@@ -37,7 +37,7 @@ function _pantheon_wordpress_update_available() {
 	include( ABSPATH . WPINC . '/version.php' );
 	
 	// Return true if our version is not the latest
-	return version_compare( $latest_wp_version, $wp_version, '>' );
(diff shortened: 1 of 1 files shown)


ataylorme:
+1, stripping `src` allows `version_compare` to work. `alpha` and `beta` can also be in the version number but are supported by `version_compare`

danielbachhuber:
@greg-1-anderson Was this included in the 4.9 release?

PR #140, opened 2017-10-31
Only show the WordPress core update nag if it's available in the Pantheon upstream

ataylorme:
There is a delay between when a WordPress release comes out and when it is merged into this repository. This results in our custom update nag being displayed during this gap and prompts users to update their upstream when the update isn't available.

pantheon-wordpress-update-nag

This pull request checks the GitHub API for a new tag in this repository to determine if a WordPress update is available, rather than the default WordPress behavior of checking WordPress.org.


ataylorme:
@greg-1-anderson happy to discuss this. I also have it applied to a multidev for review though the gap causing the issue is now over.

greg-1-anderson:
https://api.github.com is rate-limited. Do you know how often this function is called? Too many WordPress sites on one app server and we'd exceed the rate-limit, causing errors.

Maybe we could use the updatinator site as a mutex. Instead of hitting the GitHub API, curl updatinator. Then, on the uppdatinator site, we could return the cached result if available, or make a call through to the github API if we haven't done so in the last minute or so. Then we'd have up-to-date answers and no risk of going over the rate limit.

wp-content/mu-plugins/pantheon/pantheon-updates.php
```diff
@@ -3,39 +3,64 @@
...
+		if( ! is_wp_error( $response ) && 200 === $response['response']['code'] ){
+			$latest_tag_obj = end( json_decode( $response['body'] ) );
+			$pantheon_latest_wp_version = basename( $latest_tag_obj->ref );
+		} else {
+			$pantheon_latest_wp_version = $wp_version;
+		}
 
-	if( ! is_array($core_updates) || empty($core_updates) || ! property_exists($core_updates[0], 'current' ) ){
-		return null;
+		set_transient( 'pantheon_latest_wp_version', $pantheon_latest_wp_version, HOUR_IN_SECONDS );
```
ataylorme: > https://api.github.com is rate-limited. Do you know how often this function is called? Too many WordPress sites on one app server and we'd exceed the rate-limit, causing errors.

@greg-1-anderson I'm setting a WordPress transient (option that expires) stashing the Pantheon WordPress version tag from GitHub for an hour. The anonymous rate limit is 60 calls per hour. I expect ~3 per hour using the transient.

That being said caching our own value with updatinator seems like a great idea. Is that project in a state where it's able to handle such a task?

wp-content/mu-plugins/pantheon/pantheon-updates.php
```diff
@@ -3,39 +3,64 @@
...
+// Get the latest WordPress version tagged in the Pantheon upstream on GitHub
 function _pantheon_get_latest_wordpress_version() {
-	$core_updates = get_core_updates( array('dismissed' => false) );
+	$pantheon_latest_wp_version = get_transient( 'pantheon_latest_wp_version' );
+
+	if( false === $pantheon_latest_wp_version ){
+		$github_api_pantheon_upstream_tags_url = "https://api.github.com/repos/pantheon-systems/WordPress/git/refs/tags";
+		$response = wp_remote_get( $github_api_pantheon_upstream_tags_url );
+
+		if( ! is_wp_error( $response ) && 200 === $response['response']['code'] ){
```
ataylorme: Also wanted to note if we receive anything but a `200` response from GitHub the current version of WordPress the site is using is returned. This will allow things to function successfully even if rate limits are hit or GitHub returns an error.

In that scenario it would still be about an hour for the rate limit to reset and the updated value to show the nag in the dashboard. The hour delay (for the transient or rate limit reset) only apply to the WordPress dashboard nag. The prompt to update on the Pantheon dashboard will still remain.

wp-content/mu-plugins/pantheon/pantheon-updates.php
```diff
@@ -3,39 +3,64 @@
...
 }
 
-// Get the latest WordPress version
+// Get the latest WordPress version tagged in the Pantheon upstream on GitHub
 function _pantheon_get_latest_wordpress_version() {
-	$core_updates = get_core_updates( array('dismissed' => false) );
+	$pantheon_latest_wp_version = get_transient( 'pantheon_latest_wp_version' );
+
+	if( false === $pantheon_latest_wp_version ){
+		$github_api_pantheon_upstream_tags_url = "https://api.github.com/repos/pantheon-systems/WordPress/git/refs/tags";
```
ataylorme: Perhaps we should make this filterable for folks using a custom upstream.

wp-content/mu-plugins/pantheon/pantheon-updates.php
greg-1-anderson: "I don't know" is a good fallback strategy here.

wp-content/mu-plugins/pantheon/pantheon-updates.php
greg-1-anderson: I think GitHub rate limiting is per-IP address, so ~3 per hour times the number of WordPress sites running on the endpoint, plus all of the anonymous calls made by all customers using any CMS for any purpose.

If we instead called out to our own private endpoint, we could make the corresponding API call using our own credentials. Not only is the rate limit higher in this case, but more importantly, we are then in our own "pool" of requests, so we will get an answer even if some other site has exhausted the limit.

As far as Updatinator is concerned, I made a back-end site for it based on Contenta and the example-drops-8-composer repository. This would make a convenient place to add more REST endpoints for whatever purpose. This would be the first actual use, but given that the site is backed by a pull-request workflow, I think it would be maintainable.

I am hoping that at some point in the not-so-distant future to use this back end to provide other services as well.

ataylorme: @greg-1-anderson let's go for it then. Let me know how I can help. I'll refactor this PR once we have an endpoint. Happy to chip in on the updatinator side as well.

wp-content/mu-plugins/pantheon/pantheon-updates.php
greg-1-anderson: Is the custom upstream exposed by the Pantheon platform to the WordPress site? It would be better to handle this automatically for the user rather than make everyone with a custom upstream write a hook.

If we can't tell what the custom upstream is, I'd suggest that we solve this only for Pantheon's upstream for now, and leave custom upstreams as follow-on work.

ataylorme: I'm not sure if we can tell what the upstream is. I don't see it in [environment variables](https://pantheon.io/docs/read-environment-config/) and am not sure how much work it would be to add.

My concern here is if folks fork our upstream for a custom upstream they will get this `mu-plugin` and the update nag. If we can't detect the upstream we won't be able to automatically opt out for custom upstreams.

A WordPress filter would at least let them update manually without hacking the file. We could also create a filter to fire the nag at all, defaulting to `true` but allowing an override. This would be useful for sites where WordPress is installed with Composer, pushed from CI or other means besides merging the Pantheon upstream directly on the dashboard.

greg-1-anderson: Today's behavior is that users with a custom upstream will get an update nag when WordPress core has an update available, but Pantheon does not yet have a release available and therefore there is no update available for their custom upstream. In this case, they can do nothing.

With this PR, folks will not get the nag until Pantheon has a release available. If there is not yet a release available for the custom upstream, at least the user can now go and update the upstream.

If we did as you suggest and remove the nag until the custom upstream has an update available, then perhaps the utility of the nag would be eliminated, as the user might not know that they need to update the upstream.

I also see the utility in what you suggest. I think we should move forward just making this one step better (tying the nag with the availability of Pantheon's update), and then evaluate the use-case for improving the nag and how to implement it as a follow-on task. I think that most agencies are not going to want to patch their WordPress sites just to optimize the nag message; they're going to want us to do that for them. But let's discuss that separately.

greg-1-anderson:
Since the availability of the upstream update is available in the Pantheon dashboard, maybe we should just disable the nag completely, maybe replacing it with a link to the dashboard. I'm just thinking that the number of permuatations, with custom upstreams, composer-managed sites & etc. makes it imprudent to try to duplicate all of that checking here.

ataylorme:
> Since the availability of the upstream update is available in the Pantheon dashboard, maybe we should just disable the nag completely, maybe replacing it with a link to the dashboard.

The current nag does link to the dashboard. Are you saying remove the nag part and just have a dashboard link?

greg-1-anderson:
Yes, exactly - do not advertise whether or not an upgrade is available, and include the link to the dashboard invariantly.

joshkoenig:
I would support leaving it in, but possibly add some context on what to do if you don't see an update in your dashboard.

Super gold star idea: it should be possible for the environment itself to know whether it has upstream updates pending. Providing this context would allow us to deliver a much better user experience:

- If the user is on a standard upstream, but there are no updates available, tell them to check the Pantheon dashboard or contact support if they are in a hurry.
- If the user is on a custom upstream, provide that context and let them know that they may need to contact the maintainer of said upstream in order to get the update.
- If a composer upstream is detected, provide some more specific direction for that context.

ataylorme:
@joshkoenig that would be ideal but I don't believe the status of upstream updates is publicly available for us to query from WordPress. It's available via Terminus but requires authentication. If the product team is willing to work with me to make the upstream update information available to WordPress I'm happy to customize the message further.

danielbachhuber:
> that would be ideal but I don't believe the status of upstream updates is publicly available for us to query from WordPress.

Is it available through the Omnibus API (or equivalent)? See https://github.com/pantheon-systems/pantheon-hud/blob/master/inc/class-api.php#L10

ataylorme:
> Is it available through the Omnibus API (or equivalent)?

A user authenticated in WordPress is not necessarily authenticated with the Pantheon dashboard/API. Specifically this nag is helpful when folks visit the WordPress dashboard more often than the Pantheon dashboard and may miss knowing there is an upstream update available if the only notification is on the Pantheon dashboard.

danielbachhuber:
> A user authenticated in WordPress is not necessarily authenticated with the Pantheon dashboard/API.

The container can access this data though, no?

ataylorme:
I think a new `ENV` global like `PANTHEON_DEFAULT_UPSTREAM` being set to `true` or `false` would solve this but that's adding more work and would require help outside of @greg-1-anderson and myself.

joshkoenig:
@danielbachhuber is correct. The same API call a user makes to the dashboard can be done from a container with the `self` scope. 

This could be a new issue since it's more/new/different work, but basically make an omni API call to `sites/self/code-upstream-updates` and the response will contain a section like this:

```
  "dev": {
    "has_code": true,
    "is_up_to_date_with_upstream": false
  }
```

That could be used.

ataylorme:
@danielbachhuber I was able to get this working with the following:

```php
$api_url = 'https://api.live.getpantheon.com/sites/self/code-upstream-updates';
$api_response = pantheon_curl($api_url, null, 8443);
$data = json_decode($api_response['body']);
```

danielbachhuber:
Closing in favor of #212

PR #149, opened 2018-02-06
Add target="_blank" to pantheon dashboard link

IvanZugec:
It's best to add target="_blank" to the "Please update from your Pantheon dashboard." so we don't lose what's been worked on the page.


nullvariable:
Thanks for the PR! For some reason this was overlooked, but it looks like in the [more recent version](https://github.com/pantheon-systems/WordPress/blob/master/wp-content/mu-plugins/pantheon/pantheon-updates.php#L112) of this plugin that the `target="_blank"` approach has been used.

@greg-1-anderson this one can be closed.

PR #163, opened 2018-06-06
Skip code that disables update checks on test/live when called from CLI

greg-1-anderson:
… so that update checks still happen when dashboard status panel checks update status via wp-cli.

https://getpantheon.atlassian.net/browse/BUGS-1846


greg-1-anderson:
This shipped with #165

PR #167, opened 2018-07-05
Set $_SERVER['HTTPS']

### LINKED ISSUE #166: $_SERVER['HTTPS'] not set to 'on' in wp-config.php
When [`wp-config.php` checks for HTTPS](https://github.com/pantheon-systems/WordPress/blob/master/wp-config.php#L75:L77) `$_SERVER['HTTPS']` should be redefined to `on` to allow `redirect_canonical()` to function properly with HTTPS. See https://github.com/xwp/pwa-wp/issues/21 for details.

ataylorme:
Set `$_SERVER['HTTPS']` to `on` when HTTPS is detected in `wp-config.php`. Closes #166

### DIFF
--- a/wp-config.php
+++ b/wp-config.php
@@ -74,6 +74,7 @@ else:
         // content warnings.
         if (isset($_SERVER['HTTP_USER_AGENT_HTTPS']) && $_SERVER['HTTP_USER_AGENT_HTTPS'] == 'ON') {
             $scheme = 'https';
+	    $_SERVER['HTTPS'] = 'on';
         }
         define('WP_HOME', $scheme . '://' . $_SERVER['HTTP_HOST']);
         define('WP_SITEURL', $scheme . '://' . $_SERVER['HTTP_HOST']);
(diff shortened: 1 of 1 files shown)


greg-1-anderson:
Ah, you're only 1hr late for 4.9.7.

westonruter:
With this change, I believe the Pantheon docs could be updated to largely eliminate the note about modifying `wp-config.php` to manually redirect. Instead, it could advise to just change the `$scheme` from `http` to `https` so that the `WP_HOME` and `WP_SITEURL` will always use HTTPS. 

Note that an HTTPS site URL will have the effect of automatically turning on `FORCE_SSL_ADMIN`, since the constant is set to `true` in `wp_ssl_constants()` if the `siteurl` is HTTPS.

greg-1-anderson:
This PR was deployed with [WordPress 4.9.8](https://github.com/pantheon-systems/WordPress/pull/171)

PR #176, opened 2018-08-15
Default php 7.2

greg-1-anderson:


.gitignore
```diff
@@ -35,6 +35,7 @@ pantheon.upstream.yml
 *.tar
 *.zip
 *.tgz
+!wp-includes/**/*.gz
```
ataylorme: There is already a separate PR #175 for this - do we need the change duplicated here?

greg-1-anderson: @ataylorme That PR contains the same commits that I put in here. The new process for Circle testing is to merge into the `default` branch first -- the merge to master will happen automatically. We collect PRs together into a single PR to ship so that customers get just one dashboard notification on the upgrade rather than several.

ari-gold:
for commit msg, let's go with:

`Update to PHP 7.2. For details see https://pantheon.io/blog/speed-your-site-php-72/`

ari-gold:
+1 on commit msg

PR #177, opened 2018-08-23
Check that SERVER.HTTP_HOST exists before evaluating it

acobster:
Fixes a notice when run in a command-line environment, e.g. with Lando:

```
lando wp user get 1
// PHP Notice:  Undefined index: HTTP_HOST in /app/wp-content/mu-plugins/pantheon/pantheon-login-form-mods.php on line 18
```

### DIFF
--- a/.circleci/.gitignore
+++ b/.circleci/.gitignore
@@ -1,0 +1,4 @@
+.DS_Store
+local.test
+local.env
+local.ssh
--- a/.circleci/behat.yml
+++ b/.circleci/behat.yml
@@ -1,0 +1,14 @@
+# behat.yml
+default:
+  suites:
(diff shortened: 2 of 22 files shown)


acobster:
@greg-1-anderson what's the status of this? Anything I can do to help ship it?

nullvariable:
@greg-1-anderson this appears to be incorporated in the latest code at 
https://github.com/pantheon-systems/WordPress/blob/default/wp-content/mu-plugins/pantheon/pantheon-login-form-mods.php#L18

So I believe that we can close this PR

pwtyler:
Yep, looks like this was fixed in #196. Thanks, @nullvariable.

PR #180, opened 2018-10-16
Don't set "cache-control: public" for logged-in users

stevector:
@greg-1-anderson, @ataylorme, I was recently troubleshooting caching problems on a WordPress site and was surprised to see `cache-control | public, max-age=600` while looking at a logged-in response.

Our plugin sets that header outside of admin pages. That seems like a bad idea since logged-in requests to non `/admin` pages should be private too. I suspect that this issue has never been reported because the session cookie causes the behavior to be the same.

In the case of the site I was troubleshooting there were content editors who needed to to hard-refreshes of post pages after editing the posts in order to see their changes. I suspect there are more factors in play for that site.

Do you know of any reason it'd be appropriate to set this header for logged-in users? To reproduce, you should be able to look at headers for while logged into any Pantheon WordPress site.

### DIFF
--- a/wp-content/mu-plugins/pantheon/pantheon-page-cache.php
+++ b/wp-content/mu-plugins/pantheon/pantheon-page-cache.php
@@ -92,8 +92,11 @@ class Pantheon_Cache {
 
 		add_action( 'admin_post_pantheon_cache_flush_site',  array( $this, 'flush_site' ) );
 
-		if ( ! is_admin() ) {
+		if ( ! is_admin() && ! is_user_logged_in() ) {
 			add_action( 'send_headers',               array( $this, 'cache_add_headers' ) );
+		}
+		else {
+			add_action( 'send_headers',               array( $this, 'no_cache_add_headers' ) );
 		}
 		add_filter( 'rest_post_dispatch', array( $this, 'filter_rest_post_dispatch_send_cache_control' ), 10, 2 );
 
@@ -263,6 +266,17 @@ class Pantheon_Cache {
 		<?php
 	}
 
+	/**
+	 * Set a stronger cache-control header for admin or logged in requests.
+	 *
+	 * This removes "max-age=0" which could hypothetically be used by
+	 * Varnish on an immediate subsequent request.
+	 *
+	 * @return void
+	 */
+	public function no_cache_add_headers() {
+		header( 'cache-control: no-cache, no-store, must-revalidate');
+	}
 
 	/**
 	 * Add the cache-control header.
@@ -380,7 +394,7 @@ class Pantheon_Cache {
 			# If the path doesn't exist, set it to the null string
 			else {
 				$path = '';
-			}			
+			}
 			if ( '' == $path ) {
 				continue;
 			}
--- a/wp-content/mu-plugins/pantheon/pantheon-page-cache.php
+++ b/wp-content/mu-plugins/pantheon/pantheon-page-cache.php
@@ -275,7 +275,7 @@ class Pantheon_Cache {
 	 * @return void
 	 */
 	public function no_cache_add_headers() {
-		header( 'cache-control: no-cache, no-store, must-revalidate');
+		header( 'cache-control: no-cache, no-store, must-revalidate' );
 	}
 
 	/**
--- a/.circleci/.gitignore
+++ b/.circleci/.gitignore
@@ -1,0 +1,4 @@
+.DS_Store
+local.test
+local.env
+local.ssh
--- a/.circleci/behat.yml
+++ b/.circleci/behat.yml
@@ -1,0 +1,14 @@
+# behat.yml
+default:
+  suites:
+    default:
+      paths:
+        - features
+      contexts:
+        - Behat\MinkExtension\Context\MinkContext
+        - AdminLogIn
+        - ResponseHeader
+  extensions:
+    Behat\MinkExtension:
+      # base_url set by ENV
+      goutte: ~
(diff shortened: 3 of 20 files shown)


wp-content/mu-plugins/pantheon/pantheon-page-cache.php
```diff
@@ -92,7 +92,7 @@ protected function setup() {
 
 		add_action( 'admin_post_pantheon_cache_flush_site',  array( $this, 'flush_site' ) );
 
-		if ( ! is_admin() ) {
+        if ( ! is_user_logged_in() ) {
```
ataylorme: @stevector we should still check `is_admin()` so that `admin-ajax.php` calls aren't cached.

Something like `if ( ! is_admin() && ! is_user_logged_in() ) {`

davidstrauss:
I would consider explicitly adding `Cache-Control: private` for logged-in users.

stevector:
@davidstrauss, with this PR applied the header changes to `cache-control: no-cache, must-revalidate, max-age=0` Is that acceptable? I understand that `private` would be stronger.

davidstrauss:
@stevector Combining `no-cache` with `max-age` is probably inadvisable. The latter specifies aspects of the behavior forbidden by `no-cache`. I'm concerned about the resulting behavior if a browser or cache happened to treat the latter directive as an indicator that caching is actually desired. (I haven't checked the RFC to see what the behavior should be.) For example, some proxies interpret `max-age=0` to allow shared immediate use; Varnish is one such proxy.

`Cache-Control: private` is weaker than `Cache-Control: no-cache`, as the former only forbids reuse of the cache item in a shared proxy (including reverse proxies and CDNs). The latter forbids caching in the browser as well.

The best practice combination to forbid caching is `Cache-Control: no-cache, no-store, must-revalidate`.

stevector:
@davidstrauss thanks for the insight! I've updated the PR to set a stronger header.

dashifen-gu:
We ran into the hard-refresh after a page edit that brought about this patch, I think, based on an email chain between a prior employee and you all.  Due to a `.gitignore` mix up, we ended up not applying the patch to our build and wanted to let you all know that we immediately encountered that same hard-refresh need again.  We thought we'd check-in to see if this might ever be merged into the plugin itself or if we should continue to include the patch in our build process?

ataylorme:
@stevector any reason this PR includes a `.circleci` directory? If not I would like to amend it to only include the changes to `wp-content/mu-plugins/pantheon/pantheon-page-cache.php`

westonruter:
Would this explain an issue we've seen where editing a post in Gutenberg and clicking Preview does not result in the changes appearing in the Preview unless you do a hard-reload?

Video: https://youtu.be/E4zp9plVIQc

The preview document surprisingly shows as being `(from disk cache)`:

image

westonruter:
I applied this patch to the multidev environment we experienced this on, and it has fixed the issue.

stevector:
Thanks @ataylorme, I'm not sure why I started the branch to include those `.circleci` changes.

ataylorme:
It seems rebasing is conflicted. I may reset and force push this branch or start a new one

greg-1-anderson:
GitHub seems to be confused; all of those files it is showing in the .circleci directory on this branch already exist with the same content of the default branch, that this PR merges into. Maybe the branch was originally forked from master? Not sure what caused the confusion.

greg-1-anderson:
+1 on starting a new branch forked from default with just the changes you want. That should in theory clean things up.

ataylorme:
@greg-1-anderson I deleted the branch locally, re-created from `default`, and made the single file change/commit then force pushed. Should be good for review now.

Hoping to get this in for the upcoming WordPress `5.2.2` release.

greg-1-anderson:
It is difficult to get pull requests into upcoming WordPress releases.

- To make it easy: https://getpantheon.atlassian.net/browse/CORE-1187
- Failing that, make sure that a link to this PR (or any other PR desired for a release) is added to the JIRA card before that card is estimated and pulled into a sprint.

If there is no JIRA card made in advance of a release, then I'll end up doing the release as interrupt work to meet our SLA, and I won't have time to review PRs and decide if they are safe to merge into the new release.

wp-content/mu-plugins/pantheon/pantheon-page-cache.php
```diff
@@ -263,6 +266,17 @@ public function view_settings_page() {
...
+	/**
+	 * Set a stronger cache-control header for admin or logged in requests.
+	 *
+	 * This removes "max-age=0" which could hypothetically be used by
+	 * Varnish on an immediate subsequent request.
+	 *
+	 * @return void
+	 */
+	public function no_cache_add_headers() {
+		header( 'cache-control: no-cache, no-store, must-revalidate');
```
davidstrauss: It looks like there should be a corresponding space by the closing parenthesis (for symmetry).

ataylorme: Updated

ataylorme:
@greg-1-anderson with `CORE-1187` complete can this be merged into the `default` branch?

greg-1-anderson:
@ataylorme Could you re-roll so that the .circleci files are not shown as additions? Also, since "Rebase and merge" is the only option in this repository, could you please squash down to a single commit? Thanks.

ataylorme:
Closing this in favor of #213 that has a single commit and doesn't change CircleCI files.

Issue #192, opened 2019-03-13
Custom Login Screen should use customized login_message


wonkeythemonkey:
Wordpress allows you to modify the message displayed on the login screen using the `login_message` action, like so:
```php
add_filter('login_message', function () {
    return 'Login below for awesomeness!';
});
```

However, the default Wordpress login message is hard-coded into Pantheon's customizations:

https://github.com/pantheon-systems/WordPress/blob/84f059b2b4ca68ca4fa1d33dafa4eb979a542d23/wp-content/mu-plugins/pantheon/pantheon-login-form-mods.php#L60-L62

The resulting output is the custom login message above the login container and the default message inside it (see screenshot):

![Log_In_Screen_Demo](https://user-images.githubusercontent.com/4974766/54289829-3b07ad00-4580-11e9-82fa-b8438d9c893a.png)

Is it possible for Pantheon to use the customized message in place of the default message, if it exists?

Wordpress gets the custom login message using the following code:
```php
$message = apply_filters( 'login_message', $message );
if ( ! empty( $message ) ) {
    echo $message . "\n";
}
```
https://github.com/WordPress/WordPress/blob/c78d28d9f10bf9de2c2c08d0c9b529e237b6836b/wp-login.php#L202-L205

ataylorme:
Hi @wonkeythemonkey,
That seems like a reasonable request. Would you like to submit a pull request?

wonkeythemonkey:
If I can find the time in the near future, I'd love to. It will probably take a bit of time to stand up a testing environment to confirm my code works.

Do you have a contributor guide? How do you perform testing?

ataylorme:
> It will probably take a bit of time to stand up a testing environment to confirm my code works.

@wonkeythemonkey I can take care of testing if you just want to submit a PR with the code. I mostly want to make sure your name is on the PR so you get credit 😄 

If you would like to test as well the easiest way to do that would be on a Pantheon site. If you have an existing one, a [multidev environment](https://pantheon.io/docs/multidev) would work nicely. You can set it SFTP mode and make edits directly.

When the code is in a state that you like it then you can open the PR to this repository.

Once the PR is merged the change will go out to Pantheon with the next WordPress release.

PR #193, opened 2019-03-14
Use custom login_message when present

wonkeythemonkey:
Addresses https://github.com/pantheon-systems/WordPress/issues/192

If a developer has set a custom message for the login screen using `add_filter('login_message', $function);`, it should now display on the customized Pantheon login screen as well.

Merged code does the following:

* Replaces default login text with custom-set login text on the login screen
* Disables the default WP `apply_filter('login_message')` behavior to prevent the message being displayed a second time
* Slightly modifies the display CSS for the login message and Pantheon button to accommodate longer messages

### DIFF
--- a/.circleci/.gitignore
+++ b/.circleci/.gitignore
@@ -1,0 +1,4 @@
+.DS_Store
+local.test
+local.env
+local.ssh
--- a/.circleci/behat.yml
+++ b/.circleci/behat.yml
@@ -1,0 +1,14 @@
+# behat.yml
+default:
+  suites:
+    default:
+      paths:
+        - features
+      contexts:
+        - Behat\MinkExtension\Context\MinkContext
+        - AdminLogIn
+        - ResponseHeader
+  extensions:
+    Behat\MinkExtension:
+      # base_url set by ENV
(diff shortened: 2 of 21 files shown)


ataylorme:
Hi Jesse,
Thank you for the pull request! I am currently traveling for WordCamp Miami but hope to be able to test this when I return (after I catch up on email 😉 ).

ataylorme:
Screen Shot 2019-04-01 at 1 58 08 PM

wonkeythemonkey:
Seems fair enough. However, I would suggest that Pantheon's "Login to your Wordpress Site" message be hidden (or at least hideable) as the "Return to Pantheon" button is, since it is part of Pantheon's branding and not part of standard Wordpress.

wonkeythemonkey:
Also, when I set a custom message using the `login_message` filter, there is no pretty block drawn around it in Pantheon environments. It's just dumped out as plain text above the login form.

danielbachhuber:
Thanks again for your work on this, @wonkeythemonkey. I'm landing a `pantheon_wp_login_text` filter in #210

PR #196, opened 2019-03-31
Check if HTTP_HOST is set before reading it

westonruter:
I noticed an issue when invoking WP-CLI on a Lando-based Pantheon install:

```
$ lando wp plugin update query-monitor
PHP Notice:  Undefined index: HTTP_HOST in /app/wp-content/mu-plugins/pantheon/pantheon-login-form-mods.php on line 18
Notice: Undefined index: HTTP_HOST in /app/wp-content/mu-plugins/pantheon/pantheon-login-form-mods.php on line 18
```

This fixes the issue by ensuring `$_SERVER['HTTP_HOST']` is set before accessing it.


ataylorme:
@greg-1-anderson can you give a +1 to this as well?

PR #198, opened 2019-04-10
Proposal to add a Option field for REST API TTL

eabquina:
For a general use case where a customer wants to set a specific cache TTL for WP REST API endpoints, we add a specific field for the REST API TTL value

Alternatively, we can implement a custom plugin to set specific cache headers for these endpoints too, but that will be covered by another proposal or docs guide

The intention here is to avoid modifying wp-config.php to achieve this, and specifically for those who want to disable cache for

### DIFF
--- a/.circleci/.gitignore
+++ b/.circleci/.gitignore
@@ -1,0 +1,4 @@
+.DS_Store
+local.test
+local.env
+local.ssh
--- a/.circleci/behat.yml
+++ b/.circleci/behat.yml
@@ -1,0 +1,14 @@
+# behat.yml
+default:
+  suites:
+    default:
+      paths:
+        - features
+      contexts:
+        - Behat\MinkExtension\Context\MinkContext
+        - AdminLogIn
+        - ResponseHeader
+  extensions:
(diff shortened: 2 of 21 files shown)


nullvariable:
@eabquina This looks solid to me. Do other plugins like the Fastly plugin cover this feature? Is there still a need for this?

jazzsequence:
It looks like the relevant changes here are in https://github.com/pantheon-systems/WordPress/pull/198/files#diff-33a6c2a1b5ee814d965675b811abf73fc82e318914f4aef019a47d1fc47aea08 and https://github.com/pantheon-systems/WordPress/pull/198/files#diff-295abdc06a001f46eff4c4e17e8cef60b77bbf6079cfca84b68cfcb2e876cc54 -- both of which have been moved to https://github.com/pantheon-systems/pantheon-mu-plugin. This PR should be closed with the relevant changes moved over to the mu-plugin repo.

pwtyler:
Closing in favor of https://github.com/pantheon-systems/pantheon-mu-plugin/issues/24

PR #209, opened 2019-07-23
Respect custom `login_message` filter value when present

danielbachhuber:
Originally #193

If user code filters `login_message`, then the included value will display next to the "Return to Pantheon" button:

**Desktop**
image

**Mobile**
image


danielbachhuber:
D'oh, now I see @ataylorme's comment https://github.com/pantheon-systems/WordPress/pull/193#pullrequestreview-221346278

PR #213, opened 2019-09-03
Remove `max-age` from logged-in responses

ataylorme:
Use `cache-control: no-cache, no-store, must-revalidate` when on the WordPress admin and logged-in


westonruter:
It seems this didn't go far enough. I think it also needs to apply to REST API responses. See #235.

Issue #216, opened 2019-09-22
WP admin dashboard still getting update notification after updating to 5.2.3


carl-alberto:
It seems that this last update https://github.com/pantheon-systems/WordPress/commit/25bd878d315243fc5802088bf4591022e11e4bf3 is causing persistent update nags after updating WP core 5.2.3 in the following scenarios:

- Updates done from a multidev environment in the dashboard
- Accidental updates done directly from wp-admin/update-core.php https://screencast.com/t/EZxzQu2m
- Some cases of CI/CD deploys that are happening from multidev
- Possibly updates from custom upstream

It seems this line will only check the dev environment - https://github.com/pantheon-systems/WordPress/commit/25bd878d315243fc5802088bf4591022e11e4bf3?diff=unified#diff-018e375defb06713cbd1951a7dcbb9f4R25-R27

Also, the response `https://api.live.getpantheon.com/sites/self/code-upstream-updates` does not return the response for multidevs and sometimes not reflecting the correct response from `is_up_to_date_with_upstream` the api call or getting wrong transient data

# Workarounds
These workaround may work in the dev environment only after manually overwriting core via git or sftp using this method https://pantheon.io/docs/core-updates#overwrite-core:
- Clear WP transients
- Clear Redis cache
- Terminus site:upstream:clear-cache command
- Pantheon 
- Use a previous version of the pantheon-updates.php https://github.com/pantheon-systems/WordPress/blob/5.2.2/wp-content/mu-plugins/pantheon/pantheon-updates.php if verified 5.2.3 core is installed 

Maybe for the meantime, if the users are able to see the latest wp core version from the footer after doing an update, can they safely ignore the update nag @danielbachhuber @ataylorme

ataylorme:
The upstream update availability information is being pulled from the dashboard API, so incorrect values may be an issue there.

The WordPress transient only lasts for 2 minutes, so that shouldn't be much of an issue.

> Or, maybe for the meantime, if the users are able to see the latest wp core version from the footer after doing an update, can they safely ignore the update nag

The WordPress core update check is disabled. The previous version of the update notice would add the notice if WordPress thought core was out of date. However, that did not always correspond to when upstream updates were available.

For custom upstreams, updates are available more often than when WordPress core is out of date. For the default upstream, there is some delay between when a new version of WordPress is available and when the Pantheon upstream update is available.

[...]

PR #217, opened 2019-09-23
Update notification improvements

ataylorme:
Partially addresses #216 

- Do not show an update notice when using the default upstream and WordPress core is up to date
- Only show the upstream update notice on dev
  - `https://api.live.getpantheon.com/sites/self/code-upstream-updates` does not return multidev information, so displaying an accurate notice on those environments is not feasible. 
- Only disable WordPress auto updates when on Pantheon


pwtyler:
Closing in favor of #236

PR #218, opened 2019-09-30
[CORE-1353] Enforce HTTPS by default.

mbaynton:
Enforce HTTPS by default. For more info see https://pantheon.io/blog/pantheon-now-enforces-https-default-plus-really-simple-hsts

### DIFF
--- a/pantheon.upstream.yml
+++ b/pantheon.upstream.yml
@@ -4,3 +4,6 @@
 # For more information see: https://pantheon.io/docs/pantheon-upstream-yml
 api_version: 1
 php_version: 7.3
+
+# See https://pantheon.io/docs/pantheon-yml/#enforce-https--hsts for valid values.
+enforce_https: transitional


westonruter:
A couple questions about this. The docs linked above say:

> image

Is the ability to select a primary domain something available in the dashboard? I don't see it as an  option. The link to [Launch Essentials Guide](https://pantheon.io/docs/guides/launch/redirects) advises using PHP code to still do the redirect. 

Is this just not all fully rolled out yet?

greg-1-anderson:
@westonruter The documentation PR on Primary Domain is here:

https://github.com/pantheon-systems/documentation/pull/5164

I'll talk to docs about getting that merged.

greg-1-anderson:
Oh, that's one of the two documentation PRs that are being published right now.

Social promotion of these features will start at 11am; we'll have the docs updated before then.

ataylorme:
My tl;dr from that doc is that you can use `terminus domain:primary:add <site>.<env> <primary_domain>` now 😄

westonruter:
Good. I just ran that and it said it was successful. But should I not see some indication of the primary domain in the dashboard?

These links still point to the pantheonsite.io platform domain:

image

And I don't see “Primary Domain” mentioned in the list of domains on the Domains /  HTTPS page.

Lastly, when doing a database clone from live, I also still see the pantheonsite.io platform domain as the default rather than the primary domain, which I would expect.

greg-1-anderson:
The current primary domain is only usable via Terminus. We have follow-on work to bring this feature to the dashboard in the future.

PR #235, opened 2020-04-11
Ensure REST API responses are not cached for authenticated users

westonruter:
We found that that updates to posts in the block editor didn't seem to be sticking. The issue seems to be that REST API responses were being cached for logged-in users. It the change in #213 didn't go far enough. It needed to also apply to `rest_post_dispatch` and not just `send_headers` (as the latter is not triggered for REST API responses).

### DIFF
--- a/.circleci/.gitignore
+++ b/.circleci/.gitignore
@@ -1,0 +1,4 @@
+.DS_Store
+local.test
+local.env
+local.ssh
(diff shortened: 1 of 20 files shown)


westonruter:
@greg-1-anderson Can this get a review? It's still an issue. For example, we have a REST API endpoint that requires authentication and it returns with a count of posts. Currently the responses are getting cached for 10 minutes in the browser, resulting in stale data.

jspellman814:
Howdy @westonruter 👋 thanks for your contribution. I've copied your changes over to a new branch and resolved the conflicts - see PR #337. Going to close this one and work on getting the new one prepped for release.

PR #236, opened 2020-04-14
[BUGS-2808] Update notification improvements

pwtyler:
A fresh copy of @ataylorme's PR #217, as the original branch is stuck in limbo returning 404 & has merge conflict in .circleci/.

---
Partially addresses #216

- Do not show an update notice when using the default upstream and WordPress core is up to date
- Only show the upstream update notice on dev
  - https://api.live.getpantheon.com/sites/self/code-upstream-updates does not return multidev information, so displaying an accurate notice on those environments is not feasible.
- Only disable WordPress auto updates when on Pantheon

### DIFF
--- a/wp-content/mu-plugins/pantheon/pantheon-updates.php
+++ b/wp-content/mu-plugins/pantheon/pantheon-updates.php
@@ -1,12 +1,13 @@
 <?php
-// Disable WordPress auto updates
-if( ! defined('WP_AUTO_UPDATE_CORE')) {
-	define( 'WP_AUTO_UPDATE_CORE', false );
-}
-remove_action( 'wp_maybe_auto_update', 'wp_maybe_auto_update' );
-
-// Remove the default WordPress core update nag if on Pantheon
+// If on Pantheon
 if( isset( $_ENV['PANTHEON_ENVIRONMENT'] ) ){
+	// Disable WordPress auto updates
+	if( ! defined('WP_AUTO_UPDATE_CORE')) {
+		define( 'WP_AUTO_UPDATE_CORE', false );
+	}
+
+	remove_action( 'wp_maybe_auto_update', 'wp_maybe_auto_update' );
+	// Remove the default WordPress core update nag
     add_action('admin_menu','_pantheon_hide_update_nag');
 }
 
@@ -15,17 +16,70 @@ function _pantheon_hide_update_nag() {
 	remove_action( 'network_admin_notices', 'update_nag', 3 );
 }
 
-// Compare the current WordPress version to the latest available
+// Get the latest WordPress version
+function _pantheon_get_latest_wordpress_version() {
+	$core_updates = get_core_updates( array('dismissed' => false) );
+
+	if( ! is_array($core_updates) || empty($core_updates) || ! property_exists($core_updates[0], 'current' ) ){
+		return null;
+	}
+
+	return $core_updates[0]->current;
+}
+
+// Check if WordPress core is at the latest version.
+function _pantheon_is_wordpress_core_latest() {
+	$latest_wp_version = _pantheon_get_latest_wordpress_version();
+
+	if( null === $latest_wp_version ){
+		return true;
+	}
+
+	// include an unmodified $wp_version
+	include( ABSPATH . WPINC . '/version.php' );
+
+	// Return true if our version is the latest
+	return version_compare( str_replace( '-src', '', $latest_wp_version ), str_replace( '-src', '', $wp_version ), '=' );
+
+}
+
+// Return upstream's org and repository names, or false if not a custom upstream
+// (i.e. Pantheon WordPress or wordpress-network )
+function _pantheon_fetch_custom_upstream_info() {
+	$data = _pantheon_curl_cached( 'https://api.live.getpantheon.com/sites/self/code-upstream-updates' );
+	if ( empty( $data['remote_url'] ) || false !== stripos( $data['remote_url'], '/pantheon-systems/' )) {
+		// remote_url was missing or this is not a custom upstream
+		return false;
+	}
+	$url_path = ltrim( parse_url( $data['remote_url'], PHP_URL_PATH ), '/' );
+	return str_replace( '.git', '', $url_path );
+}
+
+// Check if Pantheon upstream updates are available.
 function _pantheon_wordpress_update_available() {
 
 	if ( ! function_exists( 'pantheon_curl' ) ) {
 		return false;
 	}
-	$data = _pantheon_curl_cached( 'https://api.live.getpantheon.com/sites/self/code-upstream-updates' );
-	if ( ! empty( $data['dev'] ) && isset( $data['dev']['is_up_to_date_with_upstream'] ) ) {
-		return $data['dev']['is_up_to_date_with_upstream'] ? false : true;
+
+	/**
+	 * If the site is using the default WordPress upstream and
+	 * WordPress is up to date, do not show the update notice
+	 */
+	if( ! _pantheon_fetch_custom_upstream_info() && _pantheon_is_wordpress_core_latest() ) {
+		return false;
 	}
-	return false;
+
+	$upstream_updates_api_url = 'https://api.live.getpantheon.com/sites/self/code-upstream-updates';
+	if ( 'dev' != $_ENV['PANTHEON_ENVIRONMENT'] ) {
+		$upstream_updates_api_url .= '?base_branch=refs%2Fheads%2F'.$_ENV['PANTHEON_ENVIRONMENT'];
+	}
+
+	$data = _pantheon_curl_cached( $upstream_updates_api_url );
+	if ( empty( $data['update_log'] ) ) {
+		return false;
+	}
+	return true;
 }
 
 function _pantheon_curl_cached( $api_url ) {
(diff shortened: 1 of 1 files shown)


wp-content/mu-plugins/pantheon/pantheon-updates.php
```diff
@@ -15,12 +16,54 @@ function _pantheon_hide_update_nag() {
...
+// Check if WordPress core is at the latest version.
+function _pantheon_is_wordpress_core_latest() {
+	$latest_wp_version = _pantheon_get_latest_wordpress_version();
+
+	if( null === $latest_wp_version ){
+		return true;
+	}
+
+	// include an unmodified $wp_version
+	include( ABSPATH . WPINC . '/version.php' );
```
greg-1-anderson, line 39: What context is this code being called from? It seems there should be an official way to get the current running WordPress version. Is there some reason we need to include version.php here?

wp-content/mu-plugins/pantheon/pantheon-updates.php
```diff
@@ -1,12 +1,13 @@
...
-	define( 'WP_AUTO_UPDATE_CORE', false );
-}
-remove_action( 'wp_maybe_auto_update', 'wp_maybe_auto_update' );
-
-// Remove the default WordPress core update nag if on Pantheon
+// If on Pantheon
 if( isset( $_ENV['PANTHEON_ENVIRONMENT'] ) ){
+	// Disable WordPress auto updates
+	if( ! defined('WP_AUTO_UPDATE_CORE')) {
+		define( 'WP_AUTO_UPDATE_CORE', false );
```
greg-1-anderson, line 6: +1 on doing this part (and the remove_action) only when on Pantheon.

wp-content/mu-plugins/pantheon/pantheon-updates.php
```diff
@@ -45,7 +88,7 @@ function _pantheon_upstream_update_notice() {
 	$update_type = 'new WordPress version';
 	$update_help = 'If you need help, open a support chat on Pantheon.';
 	$data = _pantheon_curl_cached( 'https://api.live.getpantheon.com/sites/self/code-upstream-updates' );
```
greg-1-anderson, line 90: Is this call to _pantheon_curl_cached superfluous? `$data` appears unused, and this same call is made again in _pantheon_is_default_upstream().

wp-content/mu-plugins/pantheon/pantheon-updates.php
```diff
@@ -45,7 +88,7 @@ function _pantheon_upstream_update_notice() {
...
-	if ( ! empty( $data['remote_url'] ) && false === stripos( $data['remote_url'], '/pantheon-systems/' ) ) {
+	if ( ! _pantheon_is_default_upstream() ) {
 		$update_type = 'Pantheon Custom Upstream update';
```
greg-1-anderson, line 92: If Pantheon *isn't* the default upstream, we change the message to 'Pantheon Custom Upstream update'? Did I read the code wrong? Seems backwards by inspection. Is the message correct when running?

wp-content/mu-plugins/pantheon/pantheon-updates.php
```diff
@@ -15,12 +16,54 @@ function _pantheon_hide_update_nag() {
 	remove_action( 'network_admin_notices', 'update_nag', 3 );
 }
 
-// Compare the current WordPress version to the latest available
+// Get the latest WordPress version
+function _pantheon_get_latest_wordpress_version() {
+	$core_updates = get_core_updates( array('dismissed' => false) );
```
greg-1-anderson, line 21: `get_core_updates` is checking to see if WordPress itself has an update, correct? I thought that the original purpose of this hook was to ensure that the nag was not displayed until the WordPress update was available on the Pantheon dashboard.

wp-content/mu-plugins/pantheon/pantheon-updates.php
pwtyler: if using the default Pantheon WordPress upstream, the message is `"new WordPress Version"`, "Pantheon Custom Upstream update" is only if the site is using a different upstream (i.e. one maintained by an org). May be more clear to have "Pantheon Upstream" vs "Pantheon Custom Upstream", but I think I prefer as written (`new WordPress version`).

wp-content/mu-plugins/pantheon/pantheon-updates.php
pwtyler: For addressing the case where core is up to date but upstream is not (ln. 63).

wp-content/mu-plugins/pantheon/pantheon-updates.php
pwtyler: Looks like it, left over from use in `if` below. I'll take it out.

wp-content/mu-plugins/pantheon/pantheon-updates.php
pwtyler: most official way would be `global $wp_version`, I think the include is to ensure that $wp_version hasn't been otherwise modified on the way here.

wp-content/mu-plugins/pantheon/pantheon-updates.php
```diff
@@ -15,12 +16,54 @@ function _pantheon_hide_update_nag() {
...
+	include( ABSPATH . WPINC . '/version.php' );
+
+	// Return true if our version is the latest
+	return version_compare( str_replace( '-src', '', $latest_wp_version ), str_replace( '-src', '', $wp_version ), '=' );
+
+}
+
+// Check if the site is using the default WordPress upstream.
+function _pantheon_is_default_upstream() {
+	$data = _pantheon_curl_cached( 'https://api.live.getpantheon.com/sites/self/code-upstream-updates' );
```
greg-1-anderson, line 48: So, here is a strategy. If we hit the above API URL, then we will always get `$data['remote_url']` and `$data['remote_head']`. These entries will be the same regardless of environment, so it does not matter that the API is always returning info for the dev branch. We can just ignore the update_log and instead use:
```
git branch --contains <sha>
* master
```
If the remote_head of our upstream is a commit in the current branch, then we are up to date, and there is no upstream update.

If remote_head does not exist in the current branch, then one of two situations is the case:

1. An upstream update is needed
2. The user updated wrong (e.g. used the WordPress dashboard, squashed multiple commits together, etc).

The second scenario will cause problems in the long run (The dashboard will think that the update is still needed, the next update won't apply, etc.), so we could consider reporting the situation and pointing the user at the right documentation page.

We can test to see if the remote has a tag for the current version via:
```
git ls-remote --tags <remote-url>
```
This will return a bunch of sha hashes followed by `refs/tags/<version>`. If there is a version that is greater than the current WordPress version installed, then we know that an update is needed. Note, however, that if the site is using a custom upstream, then the tag might not be present. Also, we might get the wrong answer if there is an upstream release that is not a WordPress release, e.g. a PHP version bump in pantheon.yml, so this check is not perfect.

pwtyler: Updated the API call to optionally include the branch if on a multidev.

wp-content/mu-plugins/pantheon/pantheon-updates.php
```diff
@@ -64,7 +111,8 @@ function _pantheon_upstream_update_notice() {
 // Register Pantheon specific WordPress update admin notice
 add_action( 'admin_init', '_pantheon_register_upstream_update_notice' );
 function _pantheon_register_upstream_update_notice(){
-    // but only if we are on Pantheon and there is a WordPress update available
+	// but only if we are on the Pantheon dev environment
```
greg-1-anderson, line 114: Only if we are on Pantheon

wp-content/mu-plugins/pantheon/pantheon-updates.php
greg-1-anderson: Could we maybe rename / refocus `_pantheon_is_default_upstream()` to return the org/project for the upstream, or `false` if the upstream is not a custom upstream (that is, if it is pantheon-systems/WordPress)? Then we could put the org/project of the custom upstream in the message. That would be a lot clearer, and very helpful. `_pantheon_is_default_upstream()` is already doing a web request that returns the remote_url.

greg-1-anderson:
Hm, the closest documentation page we have for the "bad update" problem is https://pantheon.io/docs/git-resolve-merge-conflicts.

With this PR, our handling of available updates is at least consistent with the dashboard, so maybe fixing the "bad update" problem is out of scope for this issue. Ideally we'd also fix it for the dashboard, and for Drupal and WordPress sites alike.

wp-content/mu-plugins/pantheon/pantheon-updates.php
pwtyler: Ah, yes, the ol' "changed the code but not the comments".

pwtyler:
If the "bad update" process *only* applied sanctioned core updates and nothing else, there's still a good chance one will be able to apply upstream updates and automatically resolve the merge.

> Ideally we'd also fix it for the dashboard, and for Drupal and WordPress sites alike.

+1 on keeping consistent across all cases.

wp-content/mu-plugins/pantheon/pantheon-updates.php
pwtyler: Updated function & now include org/repo in the message if using a custom upstream.

Screen Shot 2020-04-16 at 1 41 51 PM

wp-content/mu-plugins/pantheon/pantheon-updates.php
```diff
@@ -15,17 +16,73 @@ function _pantheon_hide_update_nag() {
...
+
+// Return upstream's org and repository names, or false if not a custom upstream
+// (i.e. Pantheon WordPress or wordpress-network )
+function _pantheon_fetch_custom_upstream_info() {
+	$data = _pantheon_curl_cached( 'https://api.live.getpantheon.com/sites/self/code-upstream-updates' );
+	if ( empty( $data['remote_url'] ) || false !== stripos( $data['remote_url'], '/pantheon-systems/' )) {
+		// remote_url was missing or this is not a custom upstream
+		return false;
+	}
+	$url_parts = explode('/', $data['remote_url']);
```
greg-1-anderson, line 54: Optional additional defensive coding:
- Strip of query parameters
- Strip off trailing ".git"

PR #244, opened 2020-06-15
Amended WordPress 5.4.2 release.

greg-1-anderson:
The WordPress 5.4.2 release included an additional change to the way Pantheon manages protected web paths by adding the list explicitly to the pantheon.upstream.yml file. The paths listed in the new location were missing the trailing '/' character, which changed the behavior of the path matching. This amended release restores the missing trailing slashes. No changes to WordPress itself are included.

### DIFF
--- a/pantheon.upstream.yml
+++ b/pantheon.upstream.yml
@@ -10,8 +10,8 @@ enforce_https: transitional
 
 # See https://pantheon.io/docs/pantheon-yml#protected-web-paths for usage.
 protected_web_paths:
-  - /private
-  - /wp-content/uploads/private
+  - /private/
+  - /wp-content/uploads/private/
   - /xmlrpc.php
 
 # By default, any 'protected_web_paths' added to the pantheon.yml file


brettgoldhammer:
Hi @greg-1-anderson I manage a custom upstream with about 100 production sites. After pulling this amended update into our upstream, I am still seeing some unexpected path matching throughout our environment. 

This URL was flagged to us: https://mbaexeccm.wharton.upenn.edu/private-equity and returns a 403.

It appears that despite the trailing slash in the pantheon.upstream.yml file, I see the 403 on every slug beginning with 'private' -- as an example: https://mba.wharton.upenn.edu/privateanythingafter

Can you provide any recommendations on a solution here aside from removing the /private/ line from the protected path definition?

greg-1-anderson:
@brettgoldhammer The first thing to try is to make any change (e.g. just add a comment or a blank line) to your pantheon.yml file and push it up to dev/multidev.  That will cause the private paths directives to be re-applied. I'm not sure how you could apply this update and not have that happen, but that's the first thing I would suggest.

If that does not work, copy the private web paths directive from pantheon.upstream.yml to your pantheon.yml, and set the path override parameter to `true` in your pantheon.yml.  You may remove the `/private/` entry entirely if you are not using this protected directory on your site, but it would be better to just ensure that the trailing slash is there, which should allow paths such as `private-equity` to work.

brettgoldhammer:
@greg-1-anderson Thank you for the reply -- it does seem like adding a commented line into the repo got everything to kick over and it now is functioning as expected. Much appreciated.

PR #247, opened 2020-06-27
Add a composer.json file to enable use with Composer 

mikeschinkel:
Per a discussion with Oliver Seldman he suggested using the Pantheon Upstream as our WordPress dependency instead of using [johnpbloch/wordpress-core](https://github.com/johnpbloch/wordpress-core) which is what we had been using.

Unfortunately your upstream does not contain a `composer.json` file so we cannot use it in that manner.

This PR adds the `composer.json` file that is needed to support using your upstream with Composer.

For more information about composer see [getcomposer.org](https://getcomposer.org) and [_Using Composer With WordPress_](https://www.smashingmagazine.com/2019/03/composer-wordpress/).

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -1,0 +1,37 @@
+{
+	"name": "pantheon-systems/wordpress",
+	"description": "WordPress upstream for the Pantheon website platform. Includes a platform integration plugins and a pre-configured wp-config.php.",
+	"license": "GPL-2.0-or-later",
+	"type": "wordpress-core",
+	"keywords": [
+		"wordpress",
+		"cms",
+		"webops",
+		"managed"
+	],
+	"homepage": "https://pantheon.io/",
+	"authors": [
+		{
+			"name": "WordPress Community",
+			"homepage": "https://wordpress.org/about/"
+		},
+		{
+			"name": "Pantheon",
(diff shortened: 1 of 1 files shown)


greg-1-anderson:
Quite a few years ago, Pantheon decided that it was not appropriate to add a `composer.json` to this repository, because it is missing in the upstream. I do not fully remember all of the reasons, but one may have been uncertainty around what might happen if the upstream added a `composer.json` at some later time that was in some way different or incompatible with what we did.

In any event, the reasons are immaterial. The project [pantheon-systems/wordpress-composer](https://github.com/pantheon-systems/wordpress-composer) is a derived project that maintains a copy of this repo with the addition of a composer.json file. You may use that in instances where you wish to use WordPress with Composer on Pantheon. See also [pantheon-systems/example-wordpress-composer](https://github.com/pantheon-systems/example-wordpress-composer), which provides an example of one way to use this repository.

mikeschinkel:
@greg-1-anderson — Thanks for the quick response.

> The project pantheon-systems/wordpress-composer is a derived project that maintains a copy of this repo with the addition of a composer.json file

Oh, excellent. I had not come across that. That is what we will use then. Thanks!

PR #249, opened 2020-07-23
php.ini

khalid22s-coder:
WordPress memory limit and execution time.


greg-1-anderson:
You probably intended to submit this PR against a derivative repository. However, I will also mention that you cannot change php.ini like this (or at all); the values provided for you by the platform are invariant.

PHP settings can be changed with ini_set(), for settings that can be changed by that method.

PR #245, opened 2020-06-22
Remove duplicate *.log reference.

pwtyler:
Duplicate of https://github.com/pantheon-systems/WordPress/pull/219, rebased to resolve .circleci conflict.

PR #219, opened 2019-10-10
Remove duplicate *.log reference.

kopepasah:

PR #250, opened 2020-07-30
[OTTO-395] set WP_ENVIRONMENT_TYPE according to Pantheon environment

pwtyler:
WordPress 5.5 introduces `wp_get_environment_type()`, which returns a site's environment type as defined by the constant `WP_ENVIRONMENT_TYPE`. 

The `live` environment returns `production`, `test` returns `staging`, and all other environments (`dev`, all multidevs) return `development`.

References:
https://make.wordpress.org/core/2020/07/24/new-wp_get_environment_type-function-in-wordpress-5-5/

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,7 +13,6 @@ wp-content/advanced-cache.php
 wp-content/wp-cache-config.php
 sitemap.xml
 sitemap.xml.gz
-*.log
 
 # @TODO writable paths
 wp-content/cache/
--- a/wp-config-pantheon.php
+++ b/wp-config-pantheon.php
@@ -1,0 +1,93 @@
+<?php
+/**
+ * Pantheon platform settings.
+ *
+ * IMPORTANT NOTE:
+ * Do not modify this file. This file is maintained by Pantheon.
+ *
+ * Site-specific modifications belong in wp-config.php, not this file. This
+ * file may change in future releases and modifications would cause conflicts
+ * when attempting to apply upstream updates.
+ */
+
+// ** MySQL settings - included in the Pantheon Environment ** //
+/** The name of the database for WordPress */
+define('DB_NAME', $_ENV['DB_NAME']);
+
+/** MySQL database username */
+define('DB_USER', $_ENV['DB_USER']);
+
+/** MySQL database password */
+define('DB_PASSWORD', $_ENV['DB_PASSWORD']);
+
+/** MySQL hostname; on Pantheon this includes a specific port number. */
+define('DB_HOST', $_ENV['DB_HOST'] . ':' . $_ENV['DB_PORT']);
+
+/** Database Charset to use in creating database tables. */
+define('DB_CHARSET', 'utf8mb4');
+
+/** The Database Collate type. Don't change this if in doubt. */
+define('DB_COLLATE', '');
+
+/**#@+
+ * Authentication Unique Keys and Salts.
+ *
+ * Change these to different unique phrases!
+ * You can generate these using the {@link https://api.wordpress.org/secret-key/1.1/salt/ WordPress.org secret-key service}
+ * You can change these at any point in time to invalidate all existing cookies. This will force all users to have to log in again.
+ *
+ * Pantheon sets these values for you also. If you want to shuffle them you
+ * must contact support: https://pantheon.io/docs/getting-support
+ *
+ * @since 2.6.0
+ */
+define('AUTH_KEY', $_ENV['AUTH_KEY']);
+define('SECURE_AUTH_KEY', $_ENV['SECURE_AUTH_KEY']);
+define('LOGGED_IN_KEY', $_ENV['LOGGED_IN_KEY']);
+define('NONCE_KEY', $_ENV['NONCE_KEY']);
+define('AUTH_SALT', $_ENV['AUTH_SALT']);
+define('SECURE_AUTH_SALT', $_ENV['SECURE_AUTH_SALT']);
+define('LOGGED_IN_SALT', $_ENV['LOGGED_IN_SALT']);
+define('NONCE_SALT', $_ENV['NONCE_SALT']);
+/**#@-*/
+
+/** A couple extra tweaks to help things run well on Pantheon. **/
+if (isset($_SERVER['HTTP_HOST'])) {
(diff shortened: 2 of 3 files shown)


wp-config-pantheon.php
```diff
@@ -0,0 +1,18 @@
+<?php
+
```
pwtyler, line 2: TODO: Add Comment: `DO NOT EDIT`

westonruter:
Would it not rather be preferable to specify this not in the `wp-config.php` but rather in environment variables, similar to how `PANTHEON_ENVIRONMENT` is defined? WordPress first checks `getenv( 'WP_ENVIRONMENT_TYPE' )` before checking for a `WP_ENVIRONMENT_TYPE` constant:

https://github.com/WordPress/wordpress-develop/blob/1c6ca0998c02a5c569a526324bbe2a22eb1ee9a6/src/wp-includes/load.php#L170-L181

This will reduce merge conflict conflicts with `wp-config.php`.

wp-config.php
```diff
@@ -88,7 +88,11 @@
     if ( in_array( $_ENV['PANTHEON_ENVIRONMENT'], array( 'test', 'live' ) ) && ! defined( 'DISALLOW_FILE_MODS' ) ) :
         define( 'DISALLOW_FILE_MODS', true );
     endif;
-
+    
+    if (file_exists(dirname(__FILE__) . '/wp-config-pantheon.php')) :
```
greg-1-anderson, line 92: As long as we are touching wp-config.php, let's move all of the Panthon-specific changes to the wp-config-panthon.php file. That way, if we have future need to change any Pantheon-specific behavior, we can do it without touching wp-config.php.

greg-1-anderson:
Thanks for the feedback, @westonruter. We did actually consider that strategy; however, it is more effort to do it that way, so making the change in the WordPress upstream had a number of advanteges:

1. We could guarantee that the update can be deployed before the new WordPress release with the feature comes out without sacrificing other projects.

2. Having the change in the WordPress upstream directly advertises that the feature is supported in Pantheon.

3. Making the change in WordPress now might cause merge conflicts for some people, but if we also use this opportunity to include a Pantheon-specific WordPress include file, as we should have done in the first place (and as we have done in the Drupal upstreams), then the next time we are faced with this decision, we can modify wp-config-pantheon.php and not cause merge conflicts.

wp-config-pantheon.php
```diff
@@ -0,0 +1,18 @@
...
+/**
+ * Set WP_ENVIRONMENT_TYPE according to the Pantheon Environment
+ */
+if (! defined('WP_ENVIRONMENT_TYPE')) {
```
greg-1-anderson, line 6: We should probably use getenv / putenv here.

If in the future we decided to move this logic to the prepend file, we could define WP_ENVIRONMENT in the prepend file, and that would effectively disable the code here, as intended. However, if we later decided to move this logic into environment variables, then the code here would take precedence over what we did in the platform.

While it's likely that the platform logic would be the same as what is here, we don't know if there might be some future need to do something different. So, it's a pretty minor point, but I think that putenv would be a little more flexible than define.

wp-config.php
dylanjameswagner: Is this introducing a new file named `wp-config-panthon.php`? Doesn't that open the door for moving all non-WordPress Core wp-config modifications to that file? Probably everything under the comment `/** A couple extra tweaks to help things run well on Pantheon. **/`.

greg-1-anderson: Yes, that was my suggestion exactly.

wp-config-pantheon.php
pwtyler: Overriding an ENV variable is also an option for more bespoke workflows where re-setting a constant is not, so another reason for putenv 👍

pwtyler:
Moved remainder of Pantheon-specific block into `wp-config-pantheon.php` and moved WP_ENVIRONMENT_TYPE from constants to putenv(). Added incidental change from #245 per request there. Will squash/fixup before merge once this passes further CR.

.gitignore
```diff
@@ -13,7 +13,6 @@ wp-content/advanced-cache.php
 wp-content/wp-cache-config.php
 sitemap.xml
 sitemap.xml.gz
-*.log
```
greg-1-anderson, line 16: Why was this removed? I'm neutral on this change, just wondering.

wp-config-pantheon.php
```diff
@@ -0,0 +1,87 @@
+<?php
+/**
+ * Pantheon platform settings. Everything you need should already be set.
+ * To avoid merge conflicts, no changes should be introduced to this file
```
greg-1-anderson, line 4: Don't say "to avoid merge conflicts", just say that editing is not allowed.

On the Drupal side, the comment is:
```
/**
 * @file
 * Pantheon configuration file.
 *
 * IMPORTANT NOTE:
 * Do not modify this file. This file is maintained by Pantheon.
 *
 * Site-specific modifications belong in settings.php, not this file. This file
 * may change in future releases and modifications would cause conflicts when
 * attempting to apply upstream updates.
 */
```

wp-config.php
```diff
@@ -24,71 +24,9 @@
  */
 else:
   if (isset($_ENV['PANTHEON_ENVIRONMENT'])):
```
greg-1-anderson, line 26: As long as we're modifying this section, let's switch from the odd `if ():` syntax, and use `{}` as usual here.

.gitignore
pwtyler: duplicated below (line 42). Originally opened in https://github.com/pantheon-systems/WordPress/pull/219, just been sitting for a while.

wp-config-pantheon.php
pwtyler: That was definitely a sentence I meant to circle back to (I accidentally the `.` too). Thanks.

wp-config.php
pwtyler: Already tried that— mixing the two `if` syntaxes did not play nice. We can replace *all* the if / endif throughout the file to use {} if we want, or stay consistent with if/endif— with the increased chance of conflicts (which may be a moot point given we're already expecting some).

Issue #179, opened 2018-09-04
Inconsistent indentation in the wp-config.php


dylanjameswagner:
There is mixed use of 2 space and 4 space indentation throughout the pantheon specific `wp-config.php`. This can cause confusion when adding customizations, and create a much higher chance of merge confilcts if a developer attempts to normalize the whitespace one way or the other.

Additionally, the [WordPress PHP Coding Standard](https://make.wordpress.org/core/handbook/best-practices/coding-standards/php/) is to use real tabs instead of spaces for leading whitespace. Perhaps this file should be reformatted to use the WordPress standard.

PR #253, opened 2020-08-07
[OTTO-406] wp-config.php: removed deprecated WPLANG, refactored "if" logic/syntax

### LINKED ISSUE #83: Remove WPLANG constant from wp-config.php
The `WPLANG` constant has been deprecated for quite some time. Since it's defined as an empty string, it can be safely removed.

### LINKED ISSUE #252: wp-config-pantheon.php says not to edit and to edit
`wp-config-pantheon.php` starts out by saying ([L6](https://github.com/pantheon-systems/WordPress/blob/default/wp-config-pantheon.php#L6)):
> Do not modify this file. This file is maintained by Pantheon.

Then in the hashing section, [L35](https://github.com/pantheon-systems/WordPress/blob/a9d9c261c1565670afd6a5357536c179c0b4c43e/wp-config-pantheon.php#L35) says:
> Change these to different unique phrases!

We can add information to the [documentation](https://pantheon.io/docs/migrate-manual)(update in progress to include this file) about whether and how to modify this section, and I wanted to check with you all about it first.

pwtyler:
Per https://core.trac.wordpress.org/changeset/29630

This constant was deprecated in WordPress some time ago and should be removed from wp-config altogether since it is no longer used. The database instead provides this option.

Also clarifies language around keys & salts in `wp-config-pantheon.php`

Closes #83
Closes #252

### DIFF
--- a/wp-config.php
+++ b/wp-config.php
@@ -9,47 +9,45 @@
  */
 
 /**
+ * Pantheon platform settings. Everything you need should already be set.
+ */
+if (file_exists(dirname(__FILE__) . '/wp-config-pantheon.php') && isset($_ENV['PANTHEON_ENVIRONMENT'])) {
+	require_once(dirname(__FILE__) . '/wp-config-pantheon.php');
+
+/**
  * Local configuration information.
  *
  * If you are working in a local/desktop development environment and want to
  * keep your config separate, we recommend using a 'wp-config-local.php' file,
  * which you should also make sure you .gitignore.
  */
-if (file_exists(dirname(__FILE__) . '/wp-config-local.php') && !isset($_ENV['PANTHEON_ENVIRONMENT'])):
-  # IMPORTANT: ensure your local config does not include wp-settings.php
-  require_once(dirname(__FILE__) . '/wp-config-local.php');
+} elseif (file_exists(dirname(__FILE__) . '/wp-config-local.php') && !isset($_ENV['PANTHEON_ENVIRONMENT'])){
+	# IMPORTANT: ensure your local config does not include wp-settings.php
+	require_once(dirname(__FILE__) . '/wp-config-local.php');
 
 /**
- * Pantheon platform settings. Everything you need should already be set.
+ * This block will be executed if you are NOT running on Pantheon and have NO
+ * wp-config-local.php. Insert alternate config here if necessary.
+ *
+ * If you are only running on Pantheon, you can ignore this block.
  */
-else:
-  if (isset($_ENV['PANTHEON_ENVIRONMENT'])):
-    if (file_exists(dirname(__FILE__) . '/wp-config-pantheon.php')) :
-      require_once(dirname(__FILE__) . '/wp-config-pantheon.php');
-    endif;
-  else:
-    /**
-     * This block will be executed if you have NO wp-config-local.php and you
-     * are NOT running on Pantheon. Insert alternate config here if necessary.
-     *
-     * If you are only running on Pantheon, you can ignore this block.
-     */
-    define('DB_NAME',          'database_name');
-    define('DB_USER',          'database_username');
-    define('DB_PASSWORD',      'database_password');
-    define('DB_HOST',          'database_host');
-    define('DB_CHARSET',       'utf8');
-    define('DB_COLLATE',       '');
-    define('AUTH_KEY',         'put your unique phrase here');
-    define('SECURE_AUTH_KEY',  'put your unique phrase here');
-    define('LOGGED_IN_KEY',    'put your unique phrase here');
-    define('NONCE_KEY',        'put your unique phrase here');
-    define('AUTH_SALT',        'put your unique phrase here');
-    define('SECURE_AUTH_SALT', 'put your unique phrase here');
-    define('LOGGED_IN_SALT',   'put your unique phrase here');
-    define('NONCE_SALT',       'put your unique phrase here');
-  endif;
-endif;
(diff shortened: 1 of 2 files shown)


pwtyler:
- Folding in @dylanjameswagner's proposal from #254 and fixed leading whitespace per #179 
- Updated `if/endif` -> `if{}` syntax while we're introducing the rest of these changes.

Closes #179.

wp-config.php
```diff
@@ -8,48 +8,46 @@
  * https://pantheon.io/docs
  */
 
+/**
+ * Pantheon platform settings. Everything you need should already be set.
+ */
+if (file_exists(dirname(__FILE__) . '/wp-config-pantheon.php') && isset($_ENV['PANTHEON_ENVIRONMENT'])) {
+	require_once(dirname(__FILE__) . '/wp-config-pantheon.php');
```
greg-1-anderson, line 15: nit: I think our coding guideline is two spaces.

wp-config.php
```diff
@@ -8,48 +8,46 @@
...
  * Local configuration information.
  *
  * If you are working in a local/desktop development environment and want to
  * keep your config separate, we recommend using a 'wp-config-local.php' file,
  * which you should also make sure you .gitignore.
  */
-if (file_exists(dirname(__FILE__) . '/wp-config-local.php') && !isset($_ENV['PANTHEON_ENVIRONMENT'])):
-  # IMPORTANT: ensure your local config does not include wp-settings.php
-  require_once(dirname(__FILE__) . '/wp-config-local.php');
+} elseif (file_exists(dirname(__FILE__) . '/wp-config-local.php') && !isset($_ENV['PANTHEON_ENVIRONMENT'])){
```
greg-1-anderson, line 24: You don't need both an `elseif` and the `!isset`.

I don't like else, so I'd prefer to have redundant isset / !isset.  Alternately, just skip the !isset here. On the Drupal side, the local settings file is included if present even if PANTHEON_ENVIRONMENT is set.  It's a little safer to check PANTHEON_ENVIRONMENT, though, so I'm fine with keeping that check in.

wp-config.php
```diff
@@ -8,48 +8,46 @@
...
+	# IMPORTANT: ensure your local config does not include wp-settings.php
+	require_once(dirname(__FILE__) . '/wp-config-local.php');
```
greg-1-anderson, line 26: Two spaces again

wp-config.php
```diff
@@ -8,48 +8,46 @@
...
+	define('DB_CHARSET',       'utf8');
+	define('DB_COLLATE',       '');
+	define('AUTH_KEY',         'put your unique phrase here');
+	define('SECURE_AUTH_KEY',  'put your unique phrase here');
+	define('LOGGED_IN_KEY',    'put your unique phrase here');
+	define('NONCE_KEY',        'put your unique phrase here');
+	define('AUTH_SALT',        'put your unique phrase here');
+	define('SECURE_AUTH_SALT', 'put your unique phrase here');
+	define('LOGGED_IN_SALT',   'put your unique phrase here');
+	define('NONCE_SALT',       'put your unique phrase here');
```
greg-1-anderson, line 48: Spaces, not tabs.

wp-config.php
```diff
@@ -82,7 +70,7 @@
  * "true" in dev, but false in test and live.
  */
 if ( ! defined( 'WP_DEBUG' ) ) {
-    define('WP_DEBUG', false);
+	define('WP_DEBUG', false);
```
greg-1-anderson, line 73: Two spaces, not four, not a tab.

wp-config.php
pwtyler: [WordPress Coding Standard](https://make.wordpress.org/core/handbook/best-practices/coding-standards/php/#indentation)
> Your indentation should always reflect logical structure. Use real tabs and not spaces, as this allows the most flexibility across clients.

My intention was to match WordPress standards and match `wp-config-sample.php` and keep in line with the rest of WordPress. Current wp-config has all-of-the-above.

greg-1-anderson: OK, if WP uses tabs, then it's best to stick to that. Make sure that it's tabs everywhere in the file though.

pwtyler:
Sticking with tabs per WordPress standards as mentioned in in-line comments, will squash down to a single commit then merge pending PO's final approval.

wp-config.php
```diff
@@ -8,48 +8,46 @@
...
-  # IMPORTANT: ensure your local config does not include wp-settings.php
-  require_once(dirname(__FILE__) . '/wp-config-local.php');
+} elseif (file_exists(dirname(__FILE__) . '/wp-config-local.php') && !isset($_ENV['PANTHEON_ENVIRONMENT'])){
+	# IMPORTANT: ensure your local config does not include wp-settings.php
+	require_once(dirname(__FILE__) . '/wp-config-local.php');
 
 /**
- * Pantheon platform settings. Everything you need should already be set.
+ * This block will be executed if you have NO wp-config-local.php and you
+ * are NOT running on Pantheon. Insert alternate config here if necessary.
```
dylanjameswagner, line 30: The sentence could be restructured to match the actual order the checks are done. "This block will be executed if you are NOT running on Pantheon and have NO wp-config-local.php."

pwtyler: Good call. Updated.
```
 This block will be executed if you are NOT running on Pantheon and have NO
 wp-config-local.php. Insert alternate config here if necessary.
```

PR #254, opened 2020-08-09
Refactor the logic that includes wp-config-pantheon.php

dylanjameswagner:
In PR #250 and subsequent merge commit 0331b3286e4807682ff682f29c84188c213bb92c there is the introduction of a `require_once` for the file `wp-config-pantheon.php`.

This PR refactors this logic to follow the pattern set for the `if` `file_exists` statement used for the `wp-config-local.php` and reduces the logic statements for this section to one level deep. This protects against a situation where WordPress would fail to run because the `PANTHEON_ENVIRONMENT` is set but for some reason the `wp-config-pantheon.php` file does not exist.

### DIFF
--- a/wp-config.php
+++ b/wp-config.php
@@ -9,46 +9,43 @@
  */
 
 /**
+ * Pantheon platform settings. Everything you need should already be set.
+ */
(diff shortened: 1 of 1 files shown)


pwtyler:
Thanks @dylanjameswagner, we folded this into #253— standardized spacing, and updated the if syntax from `if/endif` to `if{}` while we're making changes throughout this file anyway.

PR #261, opened 2020-09-30
improve about of wordpress .html

CLOUDROG:


nullvariable:
Hi @CLOUDROG thanks for your desire to contribute! This WordPress upstream, and specifically this file are sourced from the main WordPress repository. We don't accept changes for WordPress core files like this. You would need to make your contributions to WordPress directly. There's a [great section](https://make.wordpress.org/core/handbook/#contribute-with-code) in the contributor handbook on WordPress.org that can help you get started.

PR #264, opened 2020-10-14
Fork/174 disable update nag

nullvariable:
- Brings file inline with WordPress core coding standards
- Adds `DISABLE_PANTHEON_UPDATE_NOTICES` constant to allow disabling update nags via `wp-config.php`

PR #266, opened 2020-10-15
[BUGS-3668] Use pantheon_curl_timeout instead of pantheon_curl

pwtyler:
In addition to #263 & #264, further resolves issues with requests `api.live.getpantheon.com`.

`pantheon_curl_timeout` adds a default timeout of 1 second.


greg-1-anderson (changes requested):
Let's stop calling the API entirely, and instead append a message to the WordPress notification advising users that the Pantheon update is usually available soon after the WordPress update, and include a link to the dashboard "code" tab for the corresponding environment.

tylerdigital (approved):

pwtyler:
Following conversation with @greg-1-anderson and @tylerdigital, moving requested changes to follow-on work and merging this as-is.

PR #277, opened 2021-03-24
Send Cache Control header via Response object

### LINKED ISSUE #276: Pantheon_Cache::filter_rest_post_dispatch_send_cache_control bypasses Response abstraction
One of our customers reported an issue where a `Cannot modify header information - headers already sent by (output started at /code/wp-includes/script-loader.php:2142) in /code/wp-includes/rest-api/class-wp-rest-server.php on line 1642` warning would appear on our plugin's settings page on Pantheon.

Our plugin is preloading a REST API request using `rest_get_server()->response_to_data` which results in the `rest_post_dispatch` filter being applied because we request links to be embedded using the second parameter. Investigating the issue, it is due to the `Pantheon_Cache::filter_rest_post_dispatch_send_cache_control`. 

https://github.com/pantheon-systems/WordPress/blob/291a11331235fd454b07dda32357d904ca60b082/wp-content/mu-plugins/pantheon/pantheon-page-cache.php#L298

The cache control header is applied by sending the header directly thru `WP_REST_Server::send_header` whereas it should be applied directly to the response object using `$response->send_header`. See the `rest_send_allow_header` function for an example.

TimothyBJacobs:
Fixes #276.

### DIFF
--- a/wp-content/mu-plugins/pantheon/pantheon-page-cache.php
+++ b/wp-content/mu-plugins/pantheon/pantheon-page-cache.php
@@ -300,7 +300,7 @@ class Pantheon_Cache {
 		if ( $ttl < 60 && isset( $_ENV['PANTHEON_ENVIRONMENT'] ) && 'live' === $_ENV['PANTHEON_ENVIRONMENT'] ) {
 			$ttl = 60;
 		}
-		$server->send_header( 'Cache-Control', 'public, max-age=' . $ttl );
+		$response->header( 'Cache-Control', 'public, max-age=' . $ttl );
 		return $response;
 	}
 


pwtyler (approved):

pwtyler:
Thanks for the detailed report & fix, @TimothyBJacobs. This looks good to me pending PO 👍🏻.

nullvariable (approved):

tylerdigital:
Thanks for tracking this down and fixing @TimothyBJacobs  Your PR looks like how it should always have been done

TimothyBJacobs:
Thanks for taking a look so quickly!

PR #287, opened 2021-09-01
Restore missing update-nag formatting

### LINKED ISSUE #274: Update-nag formatting has gone missing
At some point a few months back, some of the formatting of the .update-nag was lost.

The nag used to have formatting that looked something like this:
image

And now it's missing some of the formatting that makes the notice stand out. The nag now looks like this instead:
image

If you'd agree that this nag should be formatted like other notices, then the solution might be to update this file:
wp-content/mu-plugins/pantheon/pantheon-updates.php

Where this code appears:
`<div class="update-nag">`

And update that line to include classes that apply Wordpress "notice" styling to that div, like so:
`<div class="update-nag notice notice-warning">`

Would that be okay?

joe-westcott:
Fixes #274. As noted in https://github.com/pantheon-systems/WordPress/issues/274, some of the .update-nag formatting went missing a while back, which makes the nag formatting appear somewhat inconsistent with other WordPress notices.

This commit restores the formatting that makes the nag look like other warning notices, using standard WordPress CSS class names.

### DIFF
--- a/wp-content/mu-plugins/pantheon/pantheon-updates.php
+++ b/wp-content/mu-plugins/pantheon/pantheon-updates.php
@@ -105,7 +105,7 @@ function _pantheon_upstream_update_notice() {
 	}
 
     ?>
-    <div class="update-nag">
+    <div class="update-nag notice notice-warning">
 		<p style="font-size: 14px; font-weight: bold; margin: 0 0 0.5em 0;">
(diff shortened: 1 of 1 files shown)


CdrMarks:
We had trouble getting our tests to run this PR.  Tested successfully and merged via #288.

PR #312, opened 2022-02-11
[CMS-391] Disable WP Cron

CdrMarks:
This PR disables WP Cron.  Web page requests will no long spawn requests to run wp-cron.php.  Pantheon has added support for upstreams with a framwork of `wordpress` or `wordpress-network` to automatically run cron via wp-cli once an hour.

### DIFF
--- a/wp-config-pantheon.php
+++ b/wp-config-pantheon.php
@@ -66,8 +66,6 @@ if (isset($_SERVER['HTTP_HOST'])) {
 error_reporting(E_ALL ^ E_DEPRECATED);
 /** Define appropriate location for default tmp directory on Pantheon */
 define('WP_TEMP_DIR', sys_get_temp_dir());
-/** Disable wp-cron.php from running on every page load and rely on Pantheon to run cron via wp-cli */
-define('DISABLE_WP_CRON', true);
 
 // FS writes aren't permitted in test or live, so we should let WordPress know to disable relevant UI
 if (in_array($_ENV['PANTHEON_ENVIRONMENT'], array( 'test', 'live' )) && ! defined('DISALLOW_FILE_MODS')) {
@@ -89,4 +87,15 @@ if (getenv('WP_ENVIRONMENT_TYPE') === false) {
             putenv('WP_ENVIRONMENT_TYPE=development');
             break;
     }
+}
+
+/**
+ * Defaults you may override
+ *
+ * To override, define your constant in your wp-config.php before wp-config-pantheon.php is required.
+ */
+
+/** Disable wp-cron.php from running on every page load and rely on Pantheon to run cron via wp-cli */
+if ( ! defined( 'DISABLE_WP_CRON' ) ) {
+	define( 'DISABLE_WP_CRON', true );
 }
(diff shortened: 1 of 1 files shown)


greg-1-anderson (approved):
LGTM - this file is only included when the site is running on Pantheon, so we do not need to guard individual statements.

Should probably hold off on releasing this until making a corresponding change in wp_launch_check (so that we do not report an error that cron is disabled).

westonruter:
I don't think once an hour is not frequent enough. It's true that by default `hourly` is is the lowest frequency of a scheduled event, but schedules can be added with increased frequency:

https://github.com/WordPress/wordpress-develop/blob/77232140658c01121937318f0d0c1969d6827aec/src/wp-includes/cron.php#L1045-L1073

Look at the 2,168 plugins that filter `cron_schedules`: https://wpdirectory.net/search/01FVWJTKS9ZQCAW11SWPVFXMH1

* WooCommerce and Really Simple SSL add a 1-minute interval.
* Elementor adds a 5-minute interval.
* Jetpack adds a 10-minute interval.

With the same args, schedules can be as frequent as every 10 minutes: 

https://github.com/WordPress/wordpress-develop/blob/77232140658c01121937318f0d0c1969d6827aec/src/wp-includes/cron.php#L15-L17

But this is just regarding recurring schedules. It's also common in WordPress to schedule one-off events. The most common is to schedule a post for publishing. So you could schedule 10 posts to go out over 10 minutes, with each one being published once-per-minute. On a normal WordPress install where `DISABLE_WP_CRON` is not present, sites with sufficient traffic will cause these posts to be published roughly on schedule due to cron being spawned (the caching layer notwithstanding). However, if wp-cron is scheduled only once an hour, then it's likely many scheduled posts will miss their schedules, and recurring events that are scheduled more frequently than once-an-hour will be happening much less often than expected.

I believe enabling `DISABLE_WP_CRON` should coincide with WP-CLI spawning wp-cron once a minute, not once an hour.

greg-1-anderson:
We cannot really provide a once-a-minute cron period for all bindings on the platform, at least not at this point in time. The current endpoint cron task takes more than a minute to traverse all bindings. We'd need a separate cron strategy to run cron that frequently, and not every site needs it.

Pantheon has [existing documentation on setting up an external cron service for Drupal](https://pantheon.io/docs/drupal-cron#run-cron-more-often); we will update our documentation to reflect WordPress support before we merge here.

We have also taken the measure of allowing individual sites to set DISABLE_WP_CRON to false in their wp-config.php file, before the Pantheon configuration file is loaded. If you do not want to lose the current functionality of cron running during page requests, you may set this at any time, and your site operation will not change when this is deployed.

westonruter:
Alright, but this may be considered a breaking change by sites that rely on more fine-grained scheduling.

greg-1-anderson:
Site owners will have the opportunity to read the release notes and test on the dev environment before deploying to live.

greg-1-anderson (changes requested):
The code change looks good, but CircleCI is confused and believes that the test files are being re-added. This may cause problems with the merge scripts.

Try doing a checkout -B of the PR branch based on the HEAD commit of the default branch to overwrite this branch, and cherry-pick just the desired commit. That should fix it.

greg-1-anderson (approved):

greg-1-anderson:
n.b. this will be held to go out with the next WordPress release.

PR #317, opened 2022-03-28
[FEAT-879] define('FS_METHOD', 'direct') in wp-config-pantheon.php

namespacebrian:
See https://getpantheon.atlassian.net/browse/FEAT-879

### DIFF
--- a/wp-config-pantheon.php
+++ b/wp-config-pantheon.php
@@ -92,8 +92,8 @@ if (getenv('WP_ENVIRONMENT_TYPE') === false) {
 /**
  * Resolves valhalla performance issue
  * 
- * If you see the following warning after applying this update, check if
- * you are defining this elsewhere after this code runs:
+ * If the warning below occurs, check if this constant is being defined 
+ * again in code that runs after this:
  * 
  * PHP Warning:  Constant FS_METHOD already defined in...
  */
--- a/wp-config-pantheon.php
+++ b/wp-config-pantheon.php
@@ -92,10 +92,8 @@ if (getenv('WP_ENVIRONMENT_TYPE') === false) {
 /**
  * Resolves valhalla performance issue
  * 
- * If the warning below occurs, check if this constant is being defined 
- * again in code that runs after this:
- * 
- * PHP Warning:  Constant FS_METHOD already defined in...
+ * If the site dashboard shows the error "Constant FS_METHOD already defined"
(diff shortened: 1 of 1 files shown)


wp-config-pantheon.php
```diff
@@ -88,3 +88,13 @@
             break;
     }
 }
+
+/**
+ * Resolves valhalla performance issue
+ * 
+ * If you experience dashboard errors after applying this update, check to see
```
greg-1-anderson, line 95: It won't be dashboard errors. It will be PHP errors in their site:
```
php > define('FS_METHOD', 'direct');
php > define('FS_METHOD', 'direct');
PHP Warning:  Constant FS_METHOD already defined in php shell code on line 1

Warning: Constant FS_METHOD already defined in php shell code on line 1
```
Did not run this through NGINX to see if those errors show up on the screen, if it's a WSOD, if it's in the logs, etc.

Would be good to determine that, then add a direct quote of `PHP Warning:  Constant FS_METHOD already defined` in the comment.

wp-config-pantheon.php
```diff
@@ -88,3 +88,15 @@
...
     }
 }
+
+/**
+ * Resolves valhalla performance issue
+ * 
+ * If the warning below occurs, check if this constant is being defined 
+ * again in code that runs after this:
+ * 
+ * PHP Warning:  Constant FS_METHOD already defined in...
```
greg-1-anderson, line 98: Try it with an actual site so that we can be specific about where the error message appears and what the symptoms are. Then, advise the reader to remove other occurrences of `define('FS_METHOD', ...)` from their code.

namespacebrian: When you say "where the error message appears", I assume you mean the one when the constant is defined twice, rather than the valhalla issue itself...

I will try defining it in `wp-config.php` after `wp-config-pantheon.php` is already loaded

greg-1-anderson: Yes, that is what I mean.

namespacebrian:
Error shows in site dashboard in PHP 8.0+.. and also shows up in `php-error.log`, but seems like calling out the dashboard specifically should be enough?  I didn't see it in the WP admin UI anywhere..

namespacebrian:
Closed in favor of https://github.com/pantheon-systems/WordPress/pull/319

PR #319, opened 2022-03-29
[FEAT-879= Define FS_METHOD in pantheon mu-plugin 

namespacebrian:
This setting can result in a significant performance improvement for sites that have directories with highly populated directories. This setting has been documented previously at https://pantheon.io/docs/plugins-known-issues#define-fs_method


wp-content/mu-plugins/pantheon.php
```diff
@@ -19,4 +19,7 @@
 	if ( ! defined('RETURN_TO_PANTHEON_BUTTON') || RETURN_TO_PANTHEON_BUTTON ) {
 		require_once 'pantheon/pantheon-login-form-mods.php';
 	}
+	if ( ! defined('FS_METHOD') || FS_METHOD ) {
```
jazzsequence, line 22: I'm not sure we need the `FS_METHOD` check there. This would be true if there was no defined constant OR if there WAS a defined constant (which would be incorrect).

Also, Nit: 
Typically, WP coding standards are to have spaces inside all parentheticals, so this would be:

```php
if ( ! defined( 'FS_METHOD' ) ) {
    define( 'FS_METHOD', 'direct' );
}
```

jazzsequence (changes requested):
I think bailing out of the conditional if `FS_METHOD` is truthy is wrong. Because it would be saying "if it's not defined, or if it is defined", which would be always.

wp-content/mu-plugins/pantheon.php
namespacebrian: Ah, I was trying to mimic the formatting and behaviors of the rest of the file as closely as possible, in hopes that that would be respecting coding style. 😅

I'll force-push the changes.. since I've been asked to make sure the PR is a single commit and not a stack of commits.

jazzsequence (approved):
👍

wp-content/mu-plugins/pantheon.php
jazzsequence: Yeah, the context is different for those things. 😁

namespacebrian: Yeah.. in my haste I didn't think it through.. 😳   Thank goodness for code review

namespacebrian:
Should plugin get a version bump as well?  https://github.com/pantheon-systems/WordPress/blob/default/wp-content/mu-plugins/pantheon.php#L6

greg-1-anderson:
Normally I would say "yes", but it doesn't look like the version has ever been changed. This could be the first, though. :)

greg-1-anderson:
Regarding the error where CircleCI cannot find the config, did you possibly create this PR against the `master` branch and then edit it to merge into `default`? That's my only guess as to why Circle might not have been able to find the config.

greg-1-anderson:
Tests ran here: https://github.com/pantheon-systems/WordPress/pull/314

greg-1-anderson:
I don't know that it's critical to run the tests for this change, if you did manual testing.

namespacebrian:
> Regarding the error where CircleCI cannot find the config, did you possibly create this PR against the `master` branch and then edit it to merge into `default`? That's my only guess as to why Circle might not have been able to find the config.

Uggg this is 100% what I did... the `default` branch is the default branch but I think the `master` branch is the one that drives the upstream, right?  So I need the PR to target `master`.  It just targets `default` by default and I forgot to change it.

namespacebrian:
> > Regarding the error where CircleCI cannot find the config, did you possibly create this PR against the `master` branch and then edit it to merge into `default`? That's my only guess as to why Circle might not have been able to find the config.
> 
> Uggg this is 100% what I did... the `default` branch is the default branch but I think the `master` branch is the one that drives the upstream, right? So I need the PR to target `master`. It just targets `default` by default and I forgot to change it.

4th time's the charm? :(
https://github.com/pantheon-systems/WordPress/pull/320

CdrMarks (approved):

dylanjameswagner:
Does setting FS_METHOD direct have any negative affects? I’ve only ever needed to add it temporarily when I had an issue during local dev, never for a remote. Is there a Pantheon specific reason `direct` is preferred? It’s never been set in the global config before. Shouldn’t it be set to the default with a check like how WP_DEBUG is?

Since this setting has broader affect than one service the comment should probably explain how this setting affects this platform as a whole and not just a single service.

namespacebrian:
> Does setting FS_METHOD direct have any negative affects? 

We have not identified any.

When this constant is not set, Wordpress writes and then deletes a temporary file to determine if it has direct access to the filesystem, then proceeds with the 'direct' FS_METHOD. This causes sites on the platform which write many files to encounter significantly degraded filesystem performance. In severe cases this causes degraded performance of the filesystem layer as a whole, negatively impacting the performance of other sites.

The other values for this constant--`ssh2`, `ftpext`, and `ftpsockets`--aren't ever appropriate on the platform.  So in effect, the application is multiplying its filesystem operations in order to ultimately determine what we already know: it has direct access to the filesystem.

The mu-plugin is setting the constant after all configuration files have been loaded.  If a customer were to have some need to set it to something different, they could do so in wp-config.php and that would override this change.  We chose to put it here, though, because many customers are already setting it themselves, and their code may not check to see if it is already defined, so if we deploy code that defines it before their code runs, their code would throw a warning in PHP8.

> Since this setting has broader affect than one service the comment should probably explain how this setting affects this platform as a whole and not just a single service.

I'm not sure what "a single service" refers to here.  FS_METHOD affects filesystem operations.. it doesn't affect other services like the database, DNS, etc...

dylanjameswagner:
@namespacebrian 

Thanks for the deeper explanation.

“Single service” refers to the comment in the code, `Resolves valhalla performance issue`, the single service being “valhalla”. My suggestion was to update the newly added section of code with a docblock comment that summarizes the explanation you provided rather than a general reference to a single service.

namespacebrian:
@dylanjameswagner Ah, sorry.  'Valhalla' is the name for storage layer for all site media files: `/files` which is also `wp-content/uploads` on WordPress sites and `sites/default/files` on Drupal sites.  It should be the only service affected by this, but since it's the underlying layer supporting any and all files the application can write to in live environments, it's pretty important and degraded performance affects the performance of the site (or sites) as a whole.

I see using that name is a bit too "inside baseball".  Will look at changing the comment.

jazzsequence:
> I see using that name is a bit too "inside baseball". Will look at changing the comment.

After @dylanjameswagner's comment, I agree with a general call for a more detailed doc comment. We're also defining a WordPress constant in a place where WordPress constants aren't generally defined, so a greater explanation of why we're doing that is probably in order. Something to the effect of what you said in "the application is multiplying its filesystem operations in order to ultimately determine what we already know: it has direct access to the filesystem" could probably be used. I'd recommend using a multiline comment above the `define`, rather than commenting inline.

wp-content/mu-plugins/pantheon.php
```diff
@@ -19,4 +19,16 @@
 	if ( ! defined('RETURN_TO_PANTHEON_BUTTON') || RETURN_TO_PANTHEON_BUTTON ) {
 		require_once 'pantheon/pantheon-login-form-mods.php';
 	}
+	if ( ! defined( 'FS_METHOD' ) ) {
+		/**
+		 * When this constant is not set, WordPress writes and then deletes a 
+		 * temporary file to determine if it has direct access to the filesystem,
```
jazzsequence, line 25: Would define the "it" here...

```suggestion
		 * temporary file to determine if the server has direct access to the filesystem,
```

namespacebrian: I think it == WordPress specifically

ecaron:
@namespacebrian After upgrading to PHP 8 on Pantheon, phar:///opt/pantheon/wpcli/wp-cli-2.6.0-pantheon2.phar/vendor/wp-cli/wp-cli/php/WP_CLI/Runner.php is complaining "Error: Constant FS_METHOD already defined" (discussed on internal ticket 697401).

Once solution is adjusting this `!defined` introduction to also do `!Phar::running()`. What do you think?

jazzsequence:
@ecaron If you've tested that solution and want to make a PR and/or open a new GitHub issue, I'd be happy to review. 👍

jazzsequence:
Actually, checking `defined( 'WP_CLI' ) && WP_CLI` is an established pattern in WordPress -- I haven't reproduced the issue locally, does it still work without issue if you use that instead of `!Phar::running()`?

ecaron:
I'll give that a shot. 
I liked the cleanliness of
```php
if ( ! defined( 'FS_METHOD' ) && ! Phar::running() ) {
```
but you're right that an established pattern would be better:
```php
if ( ! defined( 'FS_METHOD' ) && ( ! defined( 'WP_CLI' ) || ! WP_CLI ) ) {
```
I'll let this run a while, and if the error stays away, I'll submit a PR.

jazzsequence:
Thanks @ecaron !

PR #326, opened 2022-05-06
[CMS-142] Remove calls to Pantheon API from WordPress Admin Dashboard

jspellman814:
Verified using the following steps:

1. Created a custom upstream containing this change: https://github.com/jspellman814/custom-wp-upstream
2. Spun up a [site](https://admin.dashboard.pantheon.io/sites/54dcc744-a04e-4447-83aa-0ddfa99f7f54#dev/code) using the custom upstream.
3. Confirmed message text is shown when there is no core or upstream update available:
Screen Shot 2022-05-06 at 3 12 31 PM

4. Downgraded core to version 5.9.2
5. Confirmed message text is still shown:
core-update-needed

For follow on work, we should look into hiding this area:
core-update-needed

I don't see any hooks to do so, but we could use CSS.

### DIFF
--- a/.circleci/.gitignore
+++ b/.circleci/.gitignore
@@ -1,0 +1,4 @@
+.DS_Store
+local.test
+local.env
+local.ssh
--- a/.circleci/behat.yml
+++ b/.circleci/behat.yml
@@ -1,0 +1,14 @@
+# behat.yml
+default:
+  suites:
+    default:
+      paths:
+        - features
+      contexts:
+        - Behat\MinkExtension\Context\MinkContext
+        - AdminLogIn
+        - ResponseHeader
+  extensions:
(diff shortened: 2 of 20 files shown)


dylanjameswagner:
I’m don’t understand the issue you are trying to solve. The functionality you describe seems appropriate, if your WordPress version is out of date it makes sense that you should be notified. However, the LIVE environment doesn’t have an update available notification.

namespacebrian:
@dylanjameswagner Looking through the other bug and support tickets linked to the ticket behind this PR, it looks like all the API calls this is generating are putting excessive load on the API, which in turn slows down/causes timeouts on other operations.

jspellman814:
Closing this PR, and will open another containing the changes requested from product.

PR #330, opened 2022-05-31
[CMS-142] Remove call to Pantheon API from WP admin

jspellman814:
This PR removes the call to the Pantheon API from the WordPress admin. Most of the time these API requests are needless (e.g. there is no update available) and beyond the negative effect that has on the system as a whole, it also can cause a poor experience for the WordPress user when the API call fails under load.

With the release of this PR, the update notice behavior will be as follows:

1. If a WP Core update is detected (via [get_core_updates](https://developer.wordpress.org/reference/functions/get_core_updates/)), the update notice is shown on each admin page.
2. If a WP Core update is _not_ detected, the update notice is only shown on the `update-core` or `update-core-network` page.

The notice will still only be shown on `dev` and `multidev` environments. Users can click the notice on the updates page at any time to find out if an update is available via their Pantheon dashboard.

Notice when core is out of date:

wp59notice

Notice when core is up to date (dashboard page):

wp60notice

Notice when core is up to date (updates page):

Screen Shot 2022-05-31 at 3 21 05 PM

Old PR for reference: #326 

Edit: Notices show on dev **and** multidev environments.

### DIFF
--- a/wp-content/mu-plugins/pantheon/pantheon-updates.php
+++ b/wp-content/mu-plugins/pantheon/pantheon-updates.php
@@ -43,76 +43,45 @@ function _pantheon_is_wordpress_core_latest() {
 
 }
 
-// Return upstream's org and repository names, or false if not a custom upstream
-// (i.e. Pantheon WordPress or wordpress-network )
-function _pantheon_fetch_custom_upstream_info() {
-	$data = _pantheon_curl_cached( 'https://api.live.getpantheon.com/sites/self/code-upstream-updates' );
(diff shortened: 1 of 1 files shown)


greg-1-anderson (approved):
LGTM. In addition to the performance wins, reduced reliance on the API means more reliability, and less complexity in terms of checking for error conditions and edge cases.

pwtyler:
> The notice will still only be shown on dev environments.

Still works on multidevs as well though?

jspellman814:
> Still works on multidevs as well though?

Good catch - yeah, the notice will show on dev and multidevs. Will edit the original comment to reflect this.

jspellman814:
This will be held to go out with the next WordPress release.

PR #310, opened 2022-01-25
Behat: Look for description in the title tag

CdrMarks:
Change test since description is not visible on the page in the 2022 theme.
```
When I go to "/"                                                                         # Behat\MinkExtension\Context\MinkContext::visit()
    Then I should see "Just another WordPress site" in the ".site-description" element       # Behat\MinkExtension\Context\MinkContext::assertElementContainsText()
      Element matching css ".site-description" not found. (Behat\Mink\Exception\ElementNotFoundException)

```

PR #348, opened 2022-11-02
check for meta generator tag

jazzsequence:
All WordPress releases have the WordPress version in the meta generator tag. Since we're removing the theme stylesheet check, let's make sure that we're making sure the site is a WordPress site. 😅
We are _not_ validating the version here because that would require updating the tests with every new release.

### DIFF
--- a/.circleci/.gitignore
+++ b/.circleci/.gitignore
@@ -1,0 +1,4 @@
+.DS_Store
+local.test
+local.env
+local.ssh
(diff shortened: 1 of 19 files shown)


jspellman814:
LGTM, but don't merge this PR. Squash your commits locally, `commit --amend` your change (or rebase) into the CircleCI commit on the default branch, and then force-push it. [Linking Greg's comment](https://github.com/pantheon-systems/WordPress/pull/310#issuecomment-1021656642) regarding updates to tests for posterity.

PR #354, opened 2023-02-03
[CMSP-643] Conditionally set WP_HOME and WP_SITEURL

### LINKED ISSUE #353: Defining `WP_HOME` and `WP_SITEURL` in `wp-config.php` causes PHP notice
There are use-cases when  `WP_HOME` and `WP_SITEURL` need to be derived from an alternative header other than `HTTP_HOST`. These constants are already being defined in `wp-config-pantheon.php`. Defining these constants before or after `include 'wp-config-pantheon.php';` in `wp-config.php` will cause PHP notice.  Since `wp-config-pantheon.php` is upstream maintained, we do not recommend modifying the file due to potential merge conflict in the future. Including a modified copy of `wp-config-pantheon.php` is what we suggest but any future update will require user to manually reconcile the files. 

Proposed solution: Make the define statement in `wp-config-pantheon.php` conditional.

marfillaster:
Fix #353


wp-config-pantheon.php
```diff
@@ -59,8 +59,14 @@
         $scheme = 'https';
         $_SERVER['HTTPS'] = 'on';
     }
-    define('WP_HOME', $scheme . '://' . $_SERVER['HTTP_HOST']);
-    define('WP_SITEURL', $scheme . '://' . $_SERVER['HTTP_HOST']);
+	
+    if ( ! defined( 'WP_HOME' ) ) {
```
jazzsequence, line 63: Thanks for the PR. Not a blocker, but I prefer the "defined or define" syntax, e.g.:

```php
defined( 'WP_HOME' ) or define('WP_HOME', $scheme . '://' . $_SERVER['HTTP_HOST']);
defined( 'WP_SITEURL' ) or define('WP_SITEURL', $scheme . '://' . $_SERVER['HTTP_HOST']);
```

marfillaster: Thank you for the feedback. Applied.

jazzsequence (approved):

carl-alberto:
Newer PR that don't have a branch conflict https://github.com/pantheon-systems/WordPress/pull/433

pwtyler:
Will ship from #443.

PR #356, opened 2023-02-22
CMSP-9: Adds required constant for WordPress Multisites

rwagner00:


wp-config-pantheon.php
```diff
@@ -45,6 +45,11 @@
 define('SECURE_AUTH_SALT', $_ENV['SECURE_AUTH_SALT']);
 define('LOGGED_IN_SALT', $_ENV['LOGGED_IN_SALT']);
 define('NONCE_SALT', $_ENV['NONCE_SALT']);
+
+// When developing a WordPress Multisite locally, ensure that this constant is set.
+if (getenv('FRAMEWORK') == 'wordpress-multisite') {
```
pwtyler, line 50:
```suggestion
if (getenv('FRAMEWORK') == 'wordpress_network') {
```
Framework name is `wordpress_network`

pwtyler (changes requested):

greg-1-anderson:
I like this way better. Be sure to squash to a single commit before merging. Make the author Pantheon Automation if you want it to go straight to dashboards, or skip that step to hold until the next WordPress release.

jspellman814:
We should write [release notes](https://docs.pantheon.io/start-states/wordpress) outlining why we are making this change, and link to them from the commit message @rwagner00.

greg-1-anderson:
Looks like the tests are failing because the Pantheon fixture site needs an updated ssh key added to allow Circle to push code.

rwagner00:
https://github.com/pantheon-systems/documentation/pull/8208 is created for release notes. @jspellman814 

@greg-1-anderson is the Circle key blocking testing, or all deployment?

greg-1-anderson:
I don't think that the deploy will run unless the tests pass.

wp-config-pantheon.php
```diff
@@ -45,6 +45,12 @@
 define('SECURE_AUTH_SALT', $_ENV['SECURE_AUTH_SALT']);
 define('LOGGED_IN_SALT', $_ENV['LOGGED_IN_SALT']);
 define('NONCE_SALT', $_ENV['NONCE_SALT']);
+
+// When developing a WordPress Multisite locally, ensure that this constant is set.
+// This will set the Multisite variable in all Pantheon environments.
+if (getenv('FRAMEWORK') == 'wordpress_network') {
```
jazzsequence, line 51: Might need to add `! defined( 'WP_ALLOW_MULTISITE' ) || defined( 'WP_ALLOW_MULSTISITE' ) && ! WP_ALLOW_MULTISITE`

jazzsequence:
```suggestion
if ( getenv('FRAMEWORK') == 'wordpress_network' && ( ! defined('WP_ALLOW_MULTISITE') || defined( 'WP_ALLOW_MULTISITE' ) && ! WP_ALLOW_MULTISITE ) ) {
```

CdrMarks: When this runs, the constants that a user may define in `wp-config.php` haven't been defined yet.

pwtyler: The user could define it in wp-config before our include line, so this is worth checking and adding.

jazzsequence: Both these things are true. Ideally we'd have to advise users to update their `wp-config.php` file _or_ figure out if it's possible to declare the constant later (e.g. in the [`mu-plugin`](https://github.com/pantheon-systems/pantheon-mu-plugin/blob/main/pantheon.php) as discussed in Slack).

wp-content/mu-plugins/pantheon-mu-plugin/pantheon.php
```diff
@@ -37,4 +37,10 @@
 		 */
 		define( 'FS_METHOD', 'direct' );
 	}
+
+    // When developing a WordPress Multisite locally, ensure that this constant is set.
+    // This will set the Multisite variable in all Pantheon environments.
+    if (getenv('FRAMEWORK') == 'wordpress_network' && !defined('WP_ALLOW_MULTISITE')) {
```
jazzsequence, line 43: This file is added to the repository by updatinator, which means on a WordPress update, this will get overwritten by https://github.com/pantheon-systems/pantheon-mu-plugin. This change should go [there](https://github.com/pantheon-systems/pantheon-mu-plugin/blob/main/pantheon.php) if that's where it's going to go, and it will be updated here on a new WP release.

jazzsequence (changes requested):

rwagner00:
Closing since we're moving this change to the mu-plugin per discussion

PR #368, opened 2023-10-10
CU-85zu3jcr3 - WordPress Sentry necessary configuration has been added

cfp96:


pwtyler:
Hi @cfp96—

Plugin-specific configuration like this should be added to your individual sites' wp-config.php, or if you are using a [custom upstream](https://docs.pantheon.io/guides/custom-upstream) to manage a number of sites, wp-config.php in that repo. 

This is the default upstream for _all WordPress sites on Pantheon_.

PR #389, opened 2024-07-02
[CMSP-954] Update Default PHP to 8.2

pwtyler:
Upgrades default PHP version to 8.2. This can be overridden at the site level in `pantheon.yml`

----
https://getpantheon.atlassian.net/browse/CMSP-954


namespacebrian (approved):
Seems legit

PR #419, opened 2025-06-25
Feedback wanted: Integrate FAIR into the default WordPress upstream

jazzsequence:
This PR adds the [FAIR plugin](https://github.com/fairpm/fair-plugin) to the WordPress upstream and loads it as a mu-plugin so no configuration is required. As part of the initial integration, the default `api.wordpress.org` repository is used so the behavior of WordPress updates is retained, but this prevents many of the benefits of using the FAIR protocol including cryptographically signed package manifests that ensure plugins and themes are coming from trusted sources. As much as this PR is an actual review of the code necessary for integration, it should also be considered a starting point for discussion around FAIR broadly as well as the specific implementation. More detail below.

## Outstanding TODOs

The following work is not part of this PR but (likely) would need to be coordinated with this PR for maximum compatibility.

- [ ] **Integration with Pantheon [Update Tool](https://github.com/pantheon-systems/update-tool)** - Updatinator/Update Tool is (currently) how updates to packages (e.g. WordPress core) are managed. It is a series of scripts triggered by a CircleCI-based cron job that create the PR that will ultimately be reviewed and merged into our WordPress repository and pushed out to sites. In this case, when the WordPress.org RSS feed publishes a post announcing a new release of WordPress, the scripts pull down that latest version of WordPress, pull the latest version of the [Pantheon MU Plugin](https://github.com/pantheon-systems/pantheon-mu-plugin) and create the a PR in this repository that goes through automated testing (and internal review). **We would need to add a step to also pull down the latest version of the FAIR plugin** as a fast-follow action to merging this PR, to automate updates to the FAIR plugin, ensuring we always have the latest (or close-to-latest) version. This means that the actual version installed might lag behind updates in between WordPress releases (this is already the case for the MU plugin) as there is not currently implemented a solution that triggers a WordPress upstream update when a third party source is updated (e.g. _not_ a core update, but an update to the MU plugin or, in this case FAIR).
- [ ] **Update the default avatar setting to use [Gravatar](https://gravatar.com/)** - Out of the box, FAIR disables all specifically-Automattic-owned APIs and [replaces them with internal equivalents](https://github.com/fairpm/fair-plugin?tab=readme-ov-file#features). One of the most obvious effects of this is avatars. By disabling Gravatar and, instead allowing user-uploaded avatars, users don't need to have an account on a third-party service and their avatar image is only uploaded to the specific site. However, for sites that are currently _using_ or expecting Gravatars to be available, it will _break_ those avatars and replace them with generic defaults based on the user's display name. To make the update as seamless as possible, I propose updating the default setting to _use_ Gravatar (if unset) and allow users to explicitly _opt-out_. This work would likely need to be done in the MU plugin. See https://github.com/pantheon-systems/pantheon-mu-plugin/pull/91
- [ ] **Also integrate in [WordPress (Composer Managed)](https://github.com/pantheon-systems/wordpress-composer-managed)** - We would also want to release an update to the WordPress (Composer Managed) upstream around the same time. This update would likely be smaller, since the FAIR plugin could be added as a Composer package (**Note:** As yet, FAIR does not exist in Packagist, so a custom repository would need to be added to the WPCM `composer.json` as currently implemented.) See https://github.com/pantheon-systems/wordpress-composer-managed/pull/188

## What is FAIR?

As noted in the [FAIR project on GitHub](https://github.com/fairpm), FAIR (Federated and Independent Repositories) is a package management system that supports decentralized, cryptographically signed packages that _replaces_ the default WordPress update system (for plugins and themes only, at current release -- core support is coming). Packages are identified by a unique ID and supported by discovery servers used by the FAIR protocol. When a request to install (or update) a plugin or theme is made from inside the WordPress admin, that request queries a discovery server to find the canonical source of that package. Along the way, the protocol supports other "stops" or checkpoints, for example, there might be a moderation service that approves or rejects certain packages, or a vulnerability service that compares against a vulnerability API (like Patchstack) to see if the package and version combination is secure before the request reaches the destination repository for download/install.

By default, FAIR uses an [AspirePress](https://aspirepress.org/)-powered mirror of WordPress.org. The AspirePress mirror adds cryptographically-signed package manifests to show that they are trusted (by _AspirePress_). In the future, plugin/theme developers can host their own [mini FAIR repositories](https://github.com/fairpm/mini-fair-repo?tab=readme-ov-file) so those manifests can be signed by the developers themselves.

## FAIR on Pantheon

Our current proposal is a two-phase process. It's possible we never actually implement phase two. 

### Phase one - initial implementation

In the first phase, we hook into the FAIR protocol via the FAIR plugin. We define the `FAIR_DEFAULT_REPO_DOMAIN` to `wp-config-pantheon.php` if the constant is not set. This allows users to opt-into the mirror repository (or any other default repository) by changing this value in their own wp-config.php:

```php
define( 'FAIR_DEFAULT_REPO_DOMAIN', 'api.aspirecloud.net' );
```

**Note:** It's possible that we could define this value even _later_ in the WordPress execution by moving the constant into the MU plugin, like we do with some [other WordPress constants](https://github.com/pantheon-systems/pantheon-mu-plugin/blob/main/pantheon.php#L31-L47). That implementation detail is open to discussion.

This initial implementation also opens Pantheon users up to receiving updates from FAIR-based repositories as they become available.

### Phase two - remove the default repo override

In phase two, we would expect FAIR usage to be commonplace and that the default behavior of the FAIR plugin is sufficiently tested, supported and secure. In this case, we would entertain the possibility of _removing_ our constant definition in `wp-config-pantheon.php`. Anyone who had previously _overridden_ our override in their own `wp-config.php` file would not experience any change in behavior. If they were using a default repository _other than_ `api.aspirecloud.net`, they could leave their own constant value, otherwise they could (but would not be _required to_) remove the override.

## Try it out

I have created a fork with two branches you can use to test FAIR out on Pantheon as a custom upstream.

* **[`fair-integration`](https://github.com/jazzsequence/FAIR-WordPress/tree/fair-integration)** is the base of this PR. It implements the FAIR plugin as proposed in this PR.
* **[`fair-default`](https://github.com/jazzsequence/FAIR-WordPress/tree/fair-default)** removes the override to use `api.wordpress.org` as the `FAIR_DEFAULT_REPO_DOMAIN`.

Either branch can be used in your custom upstream. The URL to use is `https://github.com/jazzsequence/FAIR-WordPress/`.

## Considerations

**Adding things to the WordPress upstream is easy. Removing them is _hard_.** We've seen in the past artifacts left behind from previous additions to the upstream. It's easy to add new things to the WordPress upstream on Pantheon. It is distinctly more difficult to _remove_ things we've added and not cause issues in customer repositories. For this reason, integrating with FAIR is a **one-way door** and a decision that we want to be careful in deciding.

**Pantheon already protects against supply chain attacks.** In some ways, whether to use FAIR on a _platform_ level is unnecessary. One thing FAIR protects against is "calls from inside the house"-style supply chain attacks, e.g. if a user has commit access to a WordPress.org repository and decides to use that access to push out a malicious or harmful update. FAIR ensures those updates must be signed by a trusted source, otherwise they are blocked entirely from distribution. But Pantheon's dev → test → live workflow ensures that any updates coming from WordPress.org would still need to go through the same workflow, in which any issues could be identified and rolled back if necessary. Users can already install the plugin if they want to opt-into the FAIR protocol. However, we see a lot of momentum and innovation happening in this space and it reinforces our stance and existing infrastructure around supply chain security and believe that we can ease the burden of integration by building this into the platofm directly. One final point: while our workflows ultimately provide protection against supply chain attacks, we are not immune. If source code served from a heretofore trusted source (like wordpress.org) were to ship code that was significantly altered from the original state from the developers, that affects _everyone_, and is one of the primary protections a decentralized protocol provides. This is not hypothetical -- this is the case that happened a year ago with Advanced Custom Fields/Secure Custom Fields.

**Why not create a FAIR-WP upstream?** We could easily create a new Pantheon upstream and provide that to users who explicitly opt-in to use that upstream. Besides the additional maintenance burden for our engineering teams, however, this further splinters our upstream offerings. We want the experience of creating a new WordPress site to be as simple as possible. We're actively watching and participating in the conversations around FAIR and believe that _the majority of WordPress users_ will ultimately benefit from this integration.

But don't just take our word for it. **We want to hear your thoughts** about FAIR, this specific implementation and anything that is discussed above (or missing). Ask questions, leave comments or even just 👍s.

## More reading

* [FAIR discussions board](https://github.com/orgs/fairpm/discussions)
* [FAIR plugin](https://github.com/fairpm/fair-plugin)
* [FAIR protocol](https://github.com/fairpm/fair-protocol)
* [FAIR Frequently Asked Questions](https://github.com/fairpm/tsc/blob/main/faqs/README.md)
* [FAIR website](https://fair.pm)
* [FAIR 1.0 announcement post](https://fair.pm/blog/2025/09/24/discover-trust-install-fair-1-0-is-here/)
* FAIR in the news
  * [New FAIR Project Aims to Decentralize WordPress.org Services, Backed by Linux Foundation and Hundreds of Contributors](https://www.therepository.email/fair-to-decentralize-wordpress-backed-by-linux-foundation-and-contributors) (The Repository)
  * [WordPress veterans launch FAIR project to tackle security and control concerns](https://www.fastcompany.com/91347003/wordpress-veterans-launch-fair-project-to-tackle-security-and-control-concerns) (FastCompany)
  * [Linux Foundation Announces the FAIR Package Manager Project for Open Source Content Management System Stability](https://www.linuxfoundation.org/press/linux-foundation-announces-the-fair-package-manager-project-for-open-source-content-management-system-stability) (Linux Foundation)
  * [Introducing FAIR: A Stronger, More Resilient WordPress Ecosystem](https://marucchi.com/introducing-the-fair-package-manager-for-wordpress/) (Karim Marucchi)


jazzsequence:
@afragen We use a [mu-plugin loader](https://github.com/pantheon-systems/WordPress/blob/default/wp-content/mu-plugins/loader.php) in this upstream that I largely ~stole~ borrowed from Human Made. 😁 
Broadly, my goal is for users to have to know as little as possible about what FAIR is and why they should care about it; leaving it as a plugin and forcing it from your mu-loader is handy, but it would still show up on the plugins list and invite people to wonder what that is and why they can't deactivate it. Also, changing our loader at this point would likely cause merge conflicts. But this is a good suggestion to anyone watching this issue and looking for similar implementations.

jazzsequence:
Some screenshots for added context with 1.0 release:

Theme update via WP-CLI
image

Direct DID-based install screen
image

Git Updater installable via FAIR protocol
image

FAIR-hosted plugin details
image

PR #433, opened 2026-03-31
[SITE-643] Avoid the notices in WP_HOME & WP_SITE_URL

carl-alberto:
Since this has been flagged by AGCDN customers specially the one that uses domain masking that has reference to WP_HOME and WP_SITE_URL

We would like to bump this again
 
Solves the issue raised here https://github.com/pantheon-systems/WordPress/issues/353

A simpler solution from this one, I guess, since this has already been a long time https://github.com/pantheon-systems/WordPress/issues/353

### DIFF
--- a/.circleci/.gitignore
+++ b/.circleci/.gitignore
@@ -1,0 +1,4 @@
+.DS_Store
+local.test
+local.env
+local.ssh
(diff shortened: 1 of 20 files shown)


pwtyler (approved):

pwtyler:
Superseded by #443, which cherry-picks just the wp-config-pantheon.php fix onto the current default branch with an updated commit message.

